PR #21, opened 2018-01-06
change OCLExpressions to use external Type

ferdinandmehlan:
This is needed because OCL uses NumberUnit and this of now collides with the Types language here


ferdinandmehlan:
Different solution solves the problem

PR #34, opened 2022-05-30
Fix swapped argument order in GettingStarted.md

quexten:
To run the AutomataTool, the "-i" should be after the class to run, such that it is a parameter given to the main method of the AutomataTool, not a parameter given to the java runtime, which leads to an error.


quexten:
This was fixed by 1d4ced29f07d788f65e5fbbf54e164f9b670f5b1 which on the opendev branch at the time of my pull request.

N-Jansen:
Yes, the fix was part of a bigger revision of the getting started. Nevertheless, thank you for noticing and fixing this issue.

PR #67, opened 2024-12-05
Stream tc tmp

SE-FDr:
Draft until LQT.


SE-FDr:
https://github.com/MontiCore/monticore/pull/93
rebased but did not want to forcepush, thus new PR.

PR #79, opened 2025-01-14
#4448: LexerMode Parser-Attribute

maritabreuer:


monticore-generator/src/main/java/de/monticore/codegen/cd2java/_parser/ParserClassDecorator.java
```diff
@@ -79,10 +86,22 @@ protected List<ASTCDMethod> createCreateMethods(String grammarName){
...
     methods.add(createReader);
 
-
     return methods;
   }
 
+  protected List<ASTCDMember> createMode(){
+    List<ASTCDMember> members = Lists.newArrayList();
+
+    ASTCDAttribute modeAttribute = CDAttributeFacade.getInstance().createAttribute(PROTECTED.build(), "String", "mode");
```
SE-FDr, line 95: 1. would this be more readable or incorrect?
2. add small doc to setter (ALu -> is this possible if using the MutatorDecorator?) to describe what it does (and where to get the mode-String-constant from?)
```suggestion
    ASTCDAttribute modeAttribute = CDAttributeFacade.getInstance().createAttribute(PROTECTED.build(), "String", "lexerMode");
```

SE-FDr (dismissed):

PR #82, opened 2025-01-27
Deprecate (some) tc1

SE-FDr:


SE-FDr:
Merged into
https://github.com/MontiCore/monticore/pull/83

PR #94, opened 2025-02-20
Publish the reported CD/ODs of monticore-grammar

luepges:


monticore-grammar/build.gradle
```diff
@@ -575,6 +575,21 @@ if (("true").equals(getProperty('genTR'))) {
   }
 }
 
+// Pack & Publish the reported CD & OD in a jar (without the gradle module system, just a classifier)
+tasks.register('reportsJar', Jar) {
+  from (tasks.generateMCGrammars.getReportDir().dir('src/main/grammars')) {
+    include('**/*.cd', '**/*.od')
```
SE-FDr, line 581: Can this (easily) be limited to the actually reported CDs/ODs?
Thing is, what if we add our own CDs to generate Code ->
For example, one could generate the interface for the List-Fields in SymTypeExpressions,
by generating those from CDs -> so we could add CDs to mc-grammar

SE-FDr (commented):
include filter might need to be specified further?

monticore-grammar/build.gradle
luepges: Yes, this is a weird quick of Gradle syntax (used to ensure the dependency graph between tasks is correct):

we actually only pack the cd/od files within the `target/mc_reports/task_generateMCGrammars/src/main/grammars/` directory:

> Returns a [Directory](https://docs.gradle.org/current/javadoc/org/gradle/api/file/Directory.html) whose value is the given path resolved relative to the value of this directory.
https://docs.gradle.org/current/javadoc/org/gradle/api/file/DirectoryProperty.html#dir(java.lang.String)

SE-FDr (approved):

PR #108, opened 2025-03-23
Do not use src/.../resources as config templates dir unless configured

luepges:
* Do not use src/.../resources as the (config) templates dir by default for running the MontiCore Generator

Changes to a template require (as of now) a full rebuild.

Most projects do not use their own config template but contain their own templates for a generator.
If a template based generator is developed in a project, 
 the language is unnecessarily rebuild upon changes to any template.


monticore-generator/src/main/java/de/monticore/gradle/gen/MCGenTask.java
```diff
@@ -44,9 +44,9 @@ public MCGenTask() {
     // Input files to this task
     setIfPathExists(x -> this.getGrammar().from(x), Path.of(
             getProject().getProjectDir().getAbsolutePath(), "src", "main", "grammars"));
-    // Templates (such as the config templates) are loaded from there
-    setIfPathExists(this.getTmplDir()::set, Path.of(
-            getProject().getProjectDir().getAbsolutePath(), "src", "main", "resources"));
+    // Templates (such as the config templates) are NOT loaded by default,
+    // as the default usage does not use them.
+    // In case config templates are used, getTmplDir() has to be set
```
SE-FDr: Would it make sense to provide a default anyways? (!= resources folder to avoid the recompilation)

SE-FDr (approved):
Assuming pipeline runs through

PR #114, opened 2025-03-26
Check for duplicate/incorrect error codes 

luepges:
Error codes should:
 * be unique
 * start with 0x and be followed by 5-long hex number


SE-FDr (approved):
Should in theory go over all of MontCore-projects, but this alone is helpful already.

SE-FDr (approved):

PR #115, opened 2025-03-28
Cleanup Component Symbols

TomBursch:
- Cleanup DeSer
- Rename
- Use TypeCheck3 when applicable


monticore-grammar/src/main/java/de/monticore/types/check/CompKindExpression.java
```diff
@@ -43,6 +44,28 @@ public void addArgument(ASTExpression argument) {
     this.arguments.add(argument);
   }
 
+  public boolean isComponentType() {
```
schmalzing, line 51: Why do we need this? We should at least add a Docu. Not clear through its name only.

monticore-grammar/src/main/java/de/monticore/types/check/CompKindExpression.java
```diff
@@ -43,6 +44,28 @@ public void addArgument(ASTExpression argument) {
...
+    return false;
+  }
+
+  public CompKindOfComponentType asComponentType() {
+    Log.error("0xFDAB1 internal error: "
+      + "tried to convert non-component to a component."
+      + " Actual: " + this.printFullName());
+    return null;
+  }
```
schmalzing, line 64: Why do we need this? We should at least add a Docu. Not clear through its name only.

monticore-grammar/src/main/java/de/monticore/types/check/CompKindExpression.java
```diff
@@ -43,6 +44,28 @@ public void addArgument(ASTExpression argument) {
...
+
+  public boolean isGenericComponentType() {
+    return false;
+  }
```
schmalzing, line 72: Why do we need this? We should at least add a Docu. Not clear through its name only.

monticore-grammar/src/main/java/de/monticore/types/check/CompKindExpression.java
```diff
@@ -43,6 +44,28 @@ public void addArgument(ASTExpression argument) {
...
+
+  public CompKindOfGenericComponentType asGenericComponentType() {
+    Log.error("0xFDAB2 internal error: "
+      + "tried to convert non-generic-component to a generic-component."
+      + " Actual: " + this.printFullName());
+    return null;
```
schmalzing, line 82: Why do we need this? We should at least add a Docu. Not clear through its name only.

schmalzing (approved):

PR #116, opened 2025-03-31
Escape StereoValue content

luepges:


SE-FDr (approved):
Considering that the method is deprecated, I don't see why fixing it is necessary, however, now that it is implemented...

luepges:
> Considering that the method is deprecated, I don't see why fixing it is necessary, however, now that it is implemented...

That method is used (e.g. in the CD4A TOPDecorator) and the stereo-expressions addition causes illegal states due to un-escaped strings on windows machines

PR #118, opened 2025-04-01
more strict/consistent implicit conversion to String

SE-FDr:


monticore-grammar/src/main/java/de/monticore/types3/SymTypeRelations.java
```diff
@@ -253,13 +253,24 @@ public static boolean isByte(SymTypeExpression type) {
...
+   * {@link #isStringOrSubType(SymTypeExpression)} instead.
    */
   public static boolean isString(SymTypeExpression type) {
     return getDelegate()._isString(type);
   }
 
   protected abstract boolean _isString(SymTypeExpression type);
 
+  /**
+   * whether it is a String (boxed or unboxed) or a subtype (e.g., a RegEx)
```
luepges:
```suggestion
   * @param type the SymTypeExpression to check
   * @return whether it is a String (boxed or unboxed) or a subtype (e.g., a RegEx)
```

(to make the javadoc happy)

luepges (approved):

PR #120, opened 2025-04-08
Update Documentation

luepges:
* Update Getting started
* Add missing Mill JavaDocs
* Mark parser constructors as not sutiable due to missing language composition

### DIFF
--- a/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_parser/ParserClassDecorator.java
+++ b/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_parser/ParserClassDecorator.java
@@ -4,13 +4,16 @@ package de.monticore.codegen.cd2java._parser;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import de.monticore.cd.facade.CDAttributeFacade;
+import de.monticore.cd.facade.CDConstructorFacade;
 import de.monticore.cd4analysis.CD4AnalysisMill;
 import de.monticore.cd4code.CD4CodeMill;
 import de.monticore.cd4codebasis.CD4CodeBasisMill;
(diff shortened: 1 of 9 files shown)


monticore-generator/src/main/java/de/monticore/codegen/cd2java/_parser/ParserClassDecorator.java
```diff
@@ -89,6 +92,17 @@ protected List<ASTCDMethod> createCreateMethods(String grammarName){
     return methods;
   }
 
+  protected ASTCDMember createConstructor() {
+    ASTCDConstructor constr = CDConstructorFacade.getInstance().createConstructor(PUBLIC.build(), service.getParserClassSimpleName());
+    // With MC
+    this.replaceTemplate(JAVADOC, constr, JavaDoc.of("The parser for this grammar.",
+                    "{@link " + service.getMillFullName() + "#parser()} should be preferred over this constructor, as this further enables language composition.")
+            .block("deprecated", "new instances of a parser should be retrieved via a language's mill")
```
SE-FDr, line 100: Good, as it stops people (more) from using new,
but at the same time it is what we would use INSIDE the Mill, inside not deprecated methods -> so we would get warnings here, even though the Code is correct?

SE-FDr (approved):

monticore-generator/src/main/java/de/monticore/codegen/cd2java/_parser/ParserClassDecorator.java
luepges: The Mills' (generated) `_parser()` method is annotated with `SuppressWarnings("deprecation")` for this reason :) 

(I would hope that people TOPing a Mill (or modifying its templates) know what they are doing, which is why this suppression should not introduce further problems)

PR #121, opened 2025-04-09
fix check if bound is new

SE-FDr:


luepges:
Do we have a test for that (where this occurs)?

SE-FDr:
No, and we cannot, as it is only an optimization that cannot be identified from the outside (without the patch, it simply copies the constraint, which does not make sense)

luepges (approved):

PR #124, opened 2025-04-14
Add source node info to CompKindExpression

JohannesLeurs:
This will better support correct error position printing in cocos that have to do with component type checking. MontiCore provides a similar attribute for `SymTypeExpressions` via [`.getSourceInfo()`](https://github.com/MontiCore/monticore/blob/3aacee8a3791dbef6a2e14a2b7aab64424682629/monticore-grammar/src/main/java/de/monticore/types/check/SymTypeExpression.java#L691-L693). The returned object of type `SymTypeSourceInfo` has a field [`sourceNode`](https://github.com/MontiCore/monticore/blob/dev/monticore-grammar/src/main/java/de/monticore/types/check/SymTypeSourceInfo.java#L20-L24) with [associated method](https://github.com/MontiCore/monticore/blob/dev/monticore-grammar/src/main/java/de/monticore/types/check/SymTypeSourceInfo.java#L59-L72) by which we obtain the ast node on which the type expression is based, if present. 

A direct use case for having such an attribute for `CompKindExpression`s is MontiArc's type checking of generic component type arguments. E.g., the symbol table creation adds `CompKindExpression`s for declared parent components to the symbol of a component type. When we check the correct type binding of the type parameters, the entry point for the coco is the component type ast from which we obtain the component type symbol. From there on, we check all parent `CompKindExpression`s. However, there is no direct connection between the `CompKindExpression` and the ast node, in which the parent extension was declared. Thus it is hard to infer this location. The `sourceNode` attribute would help here.

### DIFF
--- a/monticore-grammar/src/main/java/de/monticore/types/check/CompKindExpression.java
+++ b/monticore-grammar/src/main/java/de/monticore/types/check/CompKindExpression.java
@@ -103,6 +103,10 @@ public abstract class CompKindExpression {
 
   /**
    * The ast node on which this CompKindExpression is based, if present.
+   * <p>
+   * This is ONLY meant to be used to create better log messages!
+   * As CompKindExpressions are moved around, cloned, and modified, it cannot be
+   * assumed that the reference to the AST node holds reliable.
    */
   public Optional<ASTNode> getSourceNode() {
     return this.sourceNode;
--- a/monticore-grammar/src/main/java/de/monticore/types/check/CompKindExpression.java
+++ b/monticore-grammar/src/main/java/de/monticore/types/check/CompKindExpression.java
@@ -2,6 +2,7 @@
 package de.monticore.types.check;
 
 import com.google.common.base.Preconditions;
+import de.monticore.ast.ASTNode;
 import de.monticore.expressions.assignmentexpressions._ast.ASTAssignmentExpression;
 import de.monticore.expressions.expressionsbasis._ast.ASTExpression;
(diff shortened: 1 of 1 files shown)


JohannesLeurs:
@schmalzing @TomBursch

JohannesLeurs:
A note about the implementation: I used `Optional<ASTNode>` as type for the reference to the ast node, while `SymTypeSourceInfo` uses [`WeakReference<ASTNode>`](https://github.com/MontiCore/monticore/blob/dev/monticore-grammar/src/main/java/de/monticore/types/check/SymTypeSourceInfo.java#L20-L24), even though the getter returns an optional. The comment attached to the weak reference says:
> Usage of a WeakReference over an Optional is mostly symbolic;
> NEVER rely on this information, but if it is available, it may be of use.

As there is no technical reason and no explanation, why `WeakReference` was chosen, I chose `Optional` instead. Okay, the use of `WeakReference` actually has technical implications: If I understand the JavaDoc correctly, the garbage collector may remove objects that are referenced by `WeakReference`s (c.f. [Java Doc: WeakReference](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/WeakReference.html)). Again, as there is no explanation for the design decision of using `WeakReference`, I chose `Optional`. However, of course, we can change this to use `WeakReference` in this case, too.

JohannesLeurs:
By the way: I suspect that the montiverse pipeline failing is caused by me not being in the MontiCore organization (it says something about missing secrets)

TomBursch (approved):

PR #126, opened 2025-04-17
Init source node of CompKindExpressions with empty optional

JohannesLeurs:
So that if `CompKindExpression#getSourceNode` is called without explicitly setting a source node or explicitly setting it absent does not return null.


JohannesLeurs:
@schmalzing @TomBursch

TomBursch (approved):

JohannesLeurs:
I will first also extend the `SynthesizeCompTypeFrom*` classes to set the `sourceNode` attribute

PR #127, opened 2025-04-17
Init CompKind  source node

JohannesLeurs:
This merge request contains two aspects:
1. Set the `CompKindExpression#sourceNode` attribute by default to an empty optional, so that calling CompKindExpression#getSourceNode can never return `null` (this was missed out in #124 
2. `SynthesizeCompTypeFrom*` sets the `sourceNode` atttribute during the synthesis of CompKindExpressions


JohannesLeurs:
@TomBursch @schmalzing

PR #131, opened 2025-04-24
Set init value for access modifiers in symbol builders correctly

JohannesLeurs:
The SymbolBuilderDecorator [contains instructions](https://github.com/MontiCore/monticore/blob/194f4409604cb883aa169f5649d3f28670ea89dc/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolBuilderDecorator.java#L145) to set an initializing value for the access modifier attribute in generated symbol builders:
```java
ASTCDAttribute accessModifier = this.getCDAttributeFacade().createAttribute(PROTECTED.build(), ACCESS_MODIFIER, "accessModifier");
this.replaceTemplate(VALUE, accessModifier, new StringHookPoint("= " + ACCESS_MODIFIER_ALL_INCLUSION));
```
However, no init value is generated for the access modifier.

The problem is the following:
1. The access modifier attribute [is created](https://github.com/MontiCore/monticore/blob/194f4409604cb883aa169f5649d3f28670ea89dc/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolBuilderDecorator.java#L58): `List<ASTCDAttribute> defaultAttrs = createSymbolAttributes(symbolClass);`
2. The access modifier attribute [is added](https://github.com/MontiCore/monticore/blob/194f4409604cb883aa169f5649d3f28670ea89dc/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolBuilderDecorator.java#L60) to the `decoratedSymbolClass` if the current symbol does not inherit it from another symbol that is extended: `decoratedSymbolClass.addAllCDMembers(defaultAttrs);`
3. The `builderDecorator` [is applied](https://github.com/MontiCore/monticore/blob/194f4409604cb883aa169f5649d3f28670ea89dc/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolBuilderDecorator.java#L74) to the `decoratedSymbolClass` in order to create a new ASTCDClass object named `symbolBuilder` which is consequently used: `ASTCDClass symbolBuilder = builderDecorator.decorate(decoratedSymbolClass);`
4. In order to do so, the `builderDecorator` [_deepClones_](https://github.com/MontiCore/monticore/blob/194f4409604cb883aa169f5649d3f28670ea89dc/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_ast/builder/BuilderDecorator.java#L69-L73) all attributes of the original CDClass
5. As a result, the `accessModifier` attribute has a differerent object identity. While we registered the template-replacement for "no initial value" to `ALL_INCLUSION` on _the original_ ASTCDAttribute, the template replacement map does not contain an entry  for _the new_ ASTCDAttribute. Hence, cd2java's default "no initial value" template is applied.

Hence, template replacements have to be transferred to during the decoration of the `BuilderDecorator`, or they have to be registered after the cloning. I did the latter in this merge request. This way, I added the replacement in the same way like it is already done for the [`build`](https://github.com/MontiCore/monticore/blob/194f4409604cb883aa169f5649d3f28670ea89dc/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolBuilderDecorator.java#L100-L105)-Method in the SymbolBuilderDecorator.


SE-FDr (approved):
Seems reasonable (not strictly the cleanest way, but surely nothing inheritly wrong with just moving everything -> that it has to be moved can be fixed later)

luepges:
(This project's github workflow setup needs some attention/does not trigger the MontiVerseCI from forks)

https://git.rwth-aachen.de/monticore/montiverseci/-/pipelines/2084150 -> passes

luepges:
Thanks :)

PR #132, opened 2025-04-25
Stereotype symbols

JohannesLeurs:
See https://git.rwth-aachen.de/monticore/monticore/-/issues/4571 :
\
The goal of this first pull request is to provide the ability of symbols to obtain stereotypes. Stereotypes themselves are considered symbols.

#### Note about design decisions:
- The field `stereoinfo` that symbols posses to record information about stereotypes they have together with associated stereovalues is named that way because of the pair-like relation between stereotypes and stereovalues: If a symbol has a stereotype with associated value, calling this pair in its entirety "stereotype" would be wrong.
- `ISymbolicStereotype extends ISymbol` is a newly introduced runtime type, which allows to add the stereoinfo attribute of type `Map<ISymbolicStereotype, Optional<Value>` to `ISymbol`. It furthermore expresses that stereotypes are symbols themselves. However, no way to resolve stereotype symbols is provided by default. If languages want to use concrete stereotype symbols, they may extend the new grammar `StereotypeSymbols`, which provides `interface symbol MCStereotype` which  implements `ISymbolicStereotype`. As a consequence of extending this grammar, the generated symboltables automatically include resolving for stereotype symbols. If the grammar is not extended, languages can still process stereotype information of symbols (e.g. when loading symbols that have stereotypes). They just can't resolve them. Thereby, languages are free to choose, if they want the resolving infrastructure, or not. In all cases, `ISymbolicStereotype` provides a common lightweight interface.

#### Discussable design points:
- The only accessor towards stereotypes that the API of `ISymbol` provides is `Map<...> getStereoinfo()`. This is opposed to, e.g., `ASTNode`. This interface provides all accessors for its `Optional<SourcePosition> start / end` and `List<pre / postComment>` that would also be generated (such as `clear_PreComments`, `add_PreComment`, `contains_PreComment`, ...). I did not add more methods, as MontiCore has only standards for generating optional and list accessors, but not map accessors. Therefore, the question is which  java.util.Map methods to mirror and how. This question becomes especially hard, as some java.util.Map methods have weird behavior when `null` becomes involved. E.g., [`Map#compute`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#compute(K,java.util.function.BiFunction)) removes a key from the map, if the computed value is `null`. It becomes weirder, when we consider that, currently, the value type of the stereoinfo map is `Optional<Value>`. On the one side, we could just adopt the semantics of java: _if the computation result is neither `Optional.of(someValue)` nor `Optional.empty()` but null, then remove the stereotype entirely from the Map)_. However, it could make more sense, if we change the semantics a bit to account for the manner of the optional value type. Take for example the method [`Map#computeIfPresent`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#computeIfPresent(K,java.util.function.BiFunction)). The java doc says
  > If the value for the specified key is present and non-null, attempts to compute a new mapping given the key and its current mapped value.

  We could nicely map this semantics to the optional manner of our value to "If the value of the optional is present, attempt to compute a new value given the key and its current mapped value". However we face a problem when mapping other parts of the specification of the method. Again we face a _remove on  null_ semantic:
  > If the remapping function returns null, the mapping is removed

  Mapping null to an empty optional, the transferred rule would be "If the mapping function returns an empty optional, the stereotype is removed from the map". Which we probably do not want, as stereotypes can also be value-less.\
  I would summarize that java's Map has weird methods. On the one hand, it allows null values for keys in maps, but on the other hand, when setting a value to null via `compute` methods, it removes the whole entry. All that we could do is to strictly adopt Java-Semantics: if the `compute` function returns `Optional.of(someValue)` or  `Optional.empty()`, no entry is removed and we can remove values from stereotypes. If the function returns `null`, the stereotype is removed from the symbol. However, this introduces some weird null handling in the code and furthermore removes most of the usefulness of functions such as `computeIfPresent` etc. (at least for our use case of optional values). I argue that introducing a generated accessor concept for `Map` attributes is out of scope of this pull request.
- As a consequence, the generated code does neither provide all the map accessors as defined in `java.util.Map`. Only a getter and setter is generated for the map. If values should be added to the stereoinfo map, then the map may  first be obtained via the getter and consequently modified. For builder classes, I added two commodity accessors: `public Builder addStereoinfo(StereoType type)` and `public Builder addStereoinfo(Stereotype type, Value value)`. I argue that this makes sense, because elsewise, while building  a symbol with stereotypes, one would have to pre-build the stereoinfo map and then call `setStereoinfo(Map ...)` on the builder. This removes the fluent-api aspect of builders, where we have an uninterrupted call chain.
- Stereotype symbols currently record the symbol types that are allowed to obtain the stereotype via a `Class<? extends ISymbol>` attribute. Having a symbolic way to refer to a symbol kind would be nice. However, symbol kinds only exist for deserialization as String constant at the moment and there is no symbolic connection between it and symbol instances.
- Stereotype symbols currently record the type of values that can be associated with them with a simple enum that mirrors the different kinds of types that the `Value` type can take and `NONE` for value-less stereotypes. For object types, this means that the type is very coarsely described as `OBJECT`. It could be interesting to describe the concrete object type to a finder degree. One could use SymTypeExpression with correspondence between `SymTypeOfPrimitive` and primitive `Value` types (_int, boolean,_ etc.) and a correspondence between `SymTypeOfObject`, `SymTypeOfGenerics` etc. with _object_ `Value` types. Another alternative is to use `Class<?>` as an attribute, as `Value`s are computed within the model processor, which, in MontiCores case, _is_ runs on the JVM → `Value` objects are JVM objects → the JVM type counterpart of JVM values are → `Class<?>`es.

#### Non goals of this pull request:
- Which accessors should be generated for `Map` attributes
- Handling of stereovalues: While the datastructure already supports their recording, nothing more is supported yet.
- How stereotype symbols are added to the symboltable (symbol files may suffice to start)
- How it is decided that a symbol obtains a symbol stereotype
- Checking symbols can obtain stereotypes only if the symbols are of compatible type; Checking that stereovalues of stereotypes have the correct type. (At the moment, it is not even decided, how symbols obtain stereotypes and associated stereovalues in the first place)
- Deserialisation of `Value`, used for symbol stereovalues. (For the moment, only stereotypes without stereovalues are supported)


monticore-grammar/src/main/grammars/de/monticore/symbols/StereotypeSymbols.mc4
```diff
@@ -0,0 +1,15 @@
...
+
+import de.monticore.*;
+
+component grammar StereotypeSymbols extends de.monticore.MCBasics {
+
+  interface symbol MCStereotype = Name;
+
+  symbolrule MCStereotype
+    implements de.monticore.symboltable.stereotypes.ISymbolicStereotype =
+      annotatedElement : Class<? extends de.monticore.symboltable.ISymbol>
```
SE-FDr: Why do we use ISymbol classes here?
If this is to specify the allowed symbol kind, those are represented by fully qualified names (s. SymbolDesers) in all other DeSerialization instances I can think of.

monticore-grammar/src/main/java/de/monticore/symbols/stereotypesymbols/_symboltable/MCStereotypeSymbolDeSer.java
```diff
@@ -0,0 +1,61 @@
...
+      s2j.getJsonPrinter().addToArray(StereoValueTypeDeSer.serializeStereoValueType(type));
+    }
+    s2j.getJsonPrinter().endArray();
+  }
+
+  @Override
+  protected Class<? extends ISymbol> deserializeAnnotatedElement(JsonObject json) {
+    String annotatedElement = json.getStringMember(ANNOTATED_ELEMENT);
+    try {
+      Class<?> clazz = Class.forName(annotatedElement);
```
SE-FDr: Yes, while technically more type safe, we would avoid reflection in this case.
Especially if a class is renamed/moved, the String representing it has to stay the same, which is not the case with this solution.

monticore-runtime/src/main/java/de/monticore/symboltable/stereotypes/ISymbolicStereotype.java
```diff
@@ -0,0 +1,13 @@
+/* (c) https://github.com/MontiCore/monticore */
+package de.monticore.symboltable.stereotypes;
+
+import de.monticore.symboltable.ISymbol;
+
+import java.util.List;
+
+public interface ISymbolicStereotype extends ISymbol {
```
SE-FDr: This class should be annotated, it is rather important

monticore-runtime/src/main/java/de/monticore/symboltable/stereotypes/StereoValueType.java
```diff
@@ -0,0 +1,14 @@
+/* (c) https://github.com/MontiCore/monticore */
+package de.monticore.symboltable.stereotypes;
+
+public enum StereoValueType {
```
SE-FDr: Should use SymTypeExpressions instead, as they can be serialized/deserialized already and can represent more Types (e.g., `Object` is not very precise)?

monticore-runtime/src/main/java/de/monticore/symboltable/stereotypes/StereoValueTypeDeSer.java
```diff
@@ -0,0 +1,74 @@
+/* (c) https://github.com/MontiCore/monticore */
+package de.monticore.symboltable.stereotypes;
+
+import de.monticore.symboltable.serialization.json.JsonElement;
+import de.monticore.symboltable.serialization.json.UserJsonString;
+import de.se_rwth.commons.logging.Log;
+
+public class StereoValueTypeDeSer {
+
+  public static final String NONE = "none";
```
SE-FDr: s. above, not needed

SE-FDr (dismissed):

monticore-runtime/src/main/java/de/monticore/symboltable/stereotypes/ISymbolicStereotype.java
JohannesLeurs: I will add documentation to the classes when it is clear what class will finally remain in the code :+1:

monticore-grammar/src/main/grammars/de/monticore/symbols/StereotypeSymbols.mc4
JohannesLeurs: I agree. See discussable points above:
> Having a symbolic way to refer to a symbol kind would be nice. However, symbol kinds only exist for deserialization as String constant at the moment and there is no symbolic connection between it and symbol instances.

However, if we only have a string, how do we "coco" check that a symbol of kind X that has the stereotype Y is allowed to have Y? We would have to draw conclusion about the _class_ that the symbol object of Y has and compare it with the string. This introduces reflection. If we do not want to base our decision on  the _class_ of a symbol, then the symbol needs a field that records which kind it has.

SE-FDr: Just read your design considerations.

> Stereotype symbols currently record the symbol types that are allowed to obtain the stereotype

This does not hold anyway IFF we consider Stereotypes on non-Symbols (which still would have an `ISymbolicStereoType`? or is that only for Symbols and others Elements would use something else?)
How would this work for, e.g., a Stereotype on an association in a CD?

> However, if we only have a string, how do we "coco" check that a symbol of kind X that has the stereotype Y is allowed to have Y?

Does this specific check even exist in the first place?
I would assume that the corresponding check would be on (Stereotype x ModelElement), not (Stereotype x Symbol).

JohannesLeurs: True, this PR only recognizes symbols. I see that we want StereotypeSymbols for AST elements, too. I had considered this out of scope for this PR, though. I will change how allowed model elements are referred to Strings for now, based on your feedback:
> If this is to specify the allowed symbol kind, those are represented by fully qualified names

Still, I am encouraging you to think about this representation: If we want to refer to certain ast nodes: how do we do that? And how do we check whether a symbol object or an ast object conforms to the model element declared as allowed by the stereotype symbol?

JohannesLeurs: I could imagine the following: Symbols and AST nodes provide a method `String getKind()`  that returns, as a String, the symbol kind / ast node kind. The _kind_ could be defined in a String constant that the instances share, omitting the need for another instance field:
```java
interface ASTNode {
  String getKind();
}

class ASTMyNode implements ASTNode {
  public static final String KIND = "foo.bar.ASTMyNode";

  @Override
  public String getKind() { return ASTMyNode.KIND; }
}
```
This solution is also easy to customize via TOP. A drawback of this method is that we can not easily check sub typing of symbols / ast nodes. As an example, imagine that we want a stereotype to be suited for _TypeSymbols_.  I guess it makes sense to allow the stereotype to also annotate _OOTypeSymbols_ then. How do we check whether `new OOTypeSymbol().getKind()` is a sub type of `new TypeSymbol().getKind()`? Maybe we can introduce a generated checking method that has a  register of sub type relations, e.g. in the type dispatcher:
```java
public boolean isSubKind(String subType, String superType) {
  // ...
  if (subType.equals(OOTypeSymbol.KIND) {
     // ...
     if (superType.equals(TypeSymbol.KIND) {
        return true;
     }
}
```
Another option to check sub type relations could  be to introduce another method in the AST / symbol classes:
```java
class ASTFoo extends ASTBar {
  @Override
  boolean isSubKindOf(String superTypeKind) {
    return superTypeKind.equals(ASTFoo.KIND)
        || superTypeKind.equals(ASTBar.KIND);
  }
}
```

A note on how to customize the kind via TOP: We just introduce a new, "overwriting" KIND constant. The getKind method and sub type checkers will adapt it, as they qualify the constant not with the TOP class name, but with the hwc class name:
```java
class ASTMyProd extends ASTMyProdTOP {
  public static final String KIND = "foo.bar.MyProdButWithChangedID";
}

class ASTMyProdTOP {
  // Will be ignored by everything referring to ASTMyProd.KIND
  public static final String KIND = "foo.bar.MyProd";

  public String getKind() {
    // Adopts an overwritten KIND  constant.
    // If the hwc class did not introduce a KIND constant, java interprets
    // the constant as ASTMCyProdTOP.KIND so we do not need to act.
    return ASTMyProd.KIND;
  }
}
```

This example used String as means to identify kinds. Of course this could also be achieved via more sophisticated means.

monticore-runtime/src/main/java/de/monticore/symboltable/stereotypes/StereoValueType.java
JohannesLeurs: Sadly, we cannot use SymTypeExpressions here. `StereoValueType` is used by `ISymbolicStereotype` which are both RTE classes. However, `SymTypeExpression` is a class in the _monticore-grammar_  project, hence part of a language component :-/

JohannesLeurs: Okay, the decision is to leave out the _allowedValueTypes_ attribute of stereotype symbols for now

monticore-grammar/src/main/grammars/de/monticore/symbols/StereotypeSymbols.mc4
SE-FDr: I am not entirely sure whether we can do this in the first place;
We cannot specify which Stereotypes a model element has during creation of the language, as there are going to be more Stereotypes that are being added.
We cannot specify which model elements are allowed to have a specific Stereotype during the creation of the Stereotype, as there might be new languages which also want to use the Stereotype as well.
We cannot easily specify the constraints on a stereotype that the model element must meet to have the stereotype, as the details of the constraints are not necessarily fully known/representable during creation of the Stereotype, wrt. adding new languages (This is only partially solved with your Kind, but not fully).
-> From what I can see, there is no restriction that we can specify that would apply generaly.

monticore-runtime/src/main/java/de/monticore/symboltable/stereotypes/ISymbolicStereotype.java
```diff
@@ -0,0 +1,24 @@
...
+ * de-/serialization capabilities.
+ *
+ * @see ISymbol#getStereoinfo()
+ */
+public interface ISymbolicStereotype extends ISymbol {
+
+  /**
+   * Defines which model elements are allowed to obtain this stereotype.
+   */
+  String getAnnotatedElement();
```
JohannesLeurs: The decision is to remove this attribute for now. It will be discussed later how to constrain which model elements may be annotated with which stereotypes

monticore-grammar/src/main/grammars/de/monticore/symbols/StereotypeSymbols.mc4
JohannesLeurs: Will be addressed in the future

JohannesLeurs: (See https://github.com/MontiCore/monticore/pull/132#discussion_r2066293052)
> The decision is to remove this attribute for now. It will be discussed later how to constrain which model elements may be annotated with which stereotypes

N-Jansen:
From my point of view, two things are off in this implementation:
1. A new grammar for introducing Stereotype symbols seems overkill. This could be part of `de.monticore.symbols.BasicSymbols`
2. I don't understand the necessity for introducing `ISymbolicStereotype` (even with the explanation above). This should be further discussed.

JohannesLeurs:
@N-Jansen 
- I renamed `ISymbolicStereotype` to `IStereotypeSymbol`
- I merged **StereotypeSymbols.mc4** into **BasicSymbols.mc4**
- I created the class [`IStereotypeReference`](https://github.com/JohannesLeurs/monticore/blob/stereotype-symbols/monticore-runtime/src/main/java/de/monticore/symboltable/stereotypes/IStereotypeReference.java) to be used in the map in [`ISymbol`](https://github.com/JohannesLeurs/monticore/blob/stereotype-symbols/monticore-runtime/src/main/java/de/monticore/symboltable/ISymbol.java#L71): `Map<IStereotypeReference, Optional<Value>> stereoinfo`. The StereotypeReference type has an  `Optional<IStereotypeSymbol> getResolved()` method. In my opinion, it is more formal than to just use the `String` name of the stereotype as key. There is the implementation [`SymbolBackedStereotypeReference`](https://github.com/JohannesLeurs/monticore/blob/stereotype-symbols/monticore-runtime/src/main/java/de/monticore/symboltable/stereotypes/SymbolBackedStereotypeReference.java) that is already available in the RTE. Moreover, I added an implementation for BasicSymbols: [`BasicSymbolsStereotypeReference`](https://github.com/JohannesLeurs/monticore/blob/stereotype-symbols/monticore-grammar/src/main/java/de/monticore/symbols/basicsymbols/_symboltable/BasicSymbolsStereotypeReference.java), which is initialized with the name for the stereotype and an enclosing scope from which the underlying stereotype symbol can be resolved. Actually, this is very similar to a Surrogate. Thus, we could also HWC extend the StereotypeSymbolSurrogate to make it implement `IStereotypeReference` and use it instead of `BasicSymbolsStereotypeReference.

What do you think about the introduction of  `IStereotypeReference` as Key type for the `stereoinfo` map in `ISymbol`?

N-Jansen (approved):

PR #156, opened 2025-06-25
Fix SymbolSurrogate spannedScope methods

TomBursch:


monticore-generator/src/main/resources/_symboltable/symbolsurrogate/SetSpannedScopeSymbolSurrogate.ftl
```diff
@@ -0,0 +1,5 @@
+<#-- (c) https://github.com/MontiCore/monticore -->
+${tc.signature("scopeName")}
+if(delegate.isPresent()){
+  lazyLoadDelegate().setSpannedScope((${scopeName}) scope);
+}
```
SE-FDr, line 5: This simply does nothing (not even Log) if no delegate has been found.
Should it not either Log.warn or at least Log.debug in this case?

monticore-generator/src/test/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolSurrogateDecoratorTest.java
```diff
@@ -209,6 +209,32 @@ public void testSetEnclosingScopeNameMethod() {
...
+
+    assertTrue(method.getMCReturnType().isPresentMCVoidType());
+
+    assertEquals(1, method.sizeCDParameters());
+    assertEquals("scope", method.getCDParameter(0).getName());
+    assertDeepEquals(I_AUTOMATON_SCOPE, method.getCDParameter(0).getMCType());
+
+    assertTrue(Log.getFindings().isEmpty());
+  }
+
```
SE-FDr, line 237: Tests for the behavior?

monticore-generator/src/main/resources/_symboltable/symbolsurrogate/SetSpannedScopeSymbolSurrogate.ftl
SE-FDr: fair, opened https://git.rwth-aachen.de/monticore/monticore/-/issues/4680 to change this in the future for all methods

monticore-generator/src/test/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolSurrogateDecoratorTest.java
SE-FDr: > AFAIK, no other of the surrogate methods are tested for behavior.
Yes, sadly...
> Is there a specific place where they should be located?
I would probably have used the BasicSymbols, as I would consider those the go-to standart, but those tests shoud be fine.

PR #162, opened 2025-07-02
Add CoCo for VarDecleration Name already defined in scope

TomBursch:


monticore-grammar/src/main/java/de/monticore/statements/mcvardeclarationstatements/_cocos/VarDeclarationNameAlreadyDefinedInScope.java
```diff
@@ -0,0 +1,41 @@
...
+   * Indicates that the name of the variable has already been used in the scope
+   */
+  public static final String ERROR_CODE = "0xA0923";
+
+  public static final String ERROR_MSG_FORMAT = "Variable '%s' is already defined in the scope.";
+
+  @Override
+  public void check(ASTVariableDeclarator node) {
+    if (!node.getDeclarator().isPresentSymbol()) {
+      Log.error(String.format("Could not find a symbol for variable '%s', thus can not check coco '%s'. Check " +
```
SE-FDr: error code?

monticore-grammar/src/main/java/de/monticore/statements/mcvardeclarationstatements/_cocos/VarDeclarationNameAlreadyDefinedInScope.java
```diff
@@ -0,0 +1,41 @@
...
+    List<VariableSymbol> localVarSymbols = node.getEnclosingScope().resolveVariableLocallyMany(
+        false,
+        node.getDeclarator().getName(),
+        AccessModifier.ALL_INCLUSION,
+        (v) -> v != node.getDeclarator().getSymbol()
+    );
+
+    if (!localVarSymbols.isEmpty()) {
+      Log.error(ERROR_CODE + " " + String.format(ERROR_MSG_FORMAT,
+          node.getDeclarator().getName()));
```
SE-FDr: SourcePosition?

monticore-grammar/src/main/java/de/monticore/statements/mcvardeclarationstatements/_cocos/VarDeclarationNameAlreadyDefinedInScope.java
SE-FDr: + SOurce Position

monticore-grammar/src/main/java/de/monticore/statements/mcvardeclarationstatements/_cocos/VarDeclarationNameAlreadyDefinedInScope.java
```diff
@@ -0,0 +1,41 @@
...
+import de.monticore.symbols.basicsymbols._symboltable.VariableSymbol;
+import de.monticore.symboltable.modifiers.AccessModifier;
+import de.se_rwth.commons.logging.Log;
+
+import java.util.List;
+
+public class VarDeclarationNameAlreadyDefinedInScope implements MCVarDeclarationStatementsASTVariableDeclaratorCoCo {
+
+  /**
+   * Indicates that the name of the variable has already been used in the scope
```
SE-FDr, line 17:
```suggestion
   * Indicates that the name of the variable has already been used in the scope
   * This will produce one error for _each_ variable (with the same name)
```

monticore-grammar/src/test/java/de/monticore/statements/cocos/VarDeclarationNameAlreadyDefinedInScopeTest.java
```diff
@@ -0,0 +1,123 @@
...
+    CombineExpressionsWithLiteralsTypeTraverserFactory.initTypeCheck3();
+    BasicSymbolsMill.initializePrimitives();
+
+    checker = new TestMCVarDeclarationStatementsCoCoChecker();
+    checker.setTraverser(TestMCVarDeclarationStatementsMill.traverser());
+    checker.addCoCo(new VarDeclarationNameAlreadyDefinedInScope());
+    parser = TestMCVarDeclarationStatementsMill.parser();
+  }
+
+  protected void checkExpectedErrorsPresent(List<String> expectedErrorCodes) {
```
SE-FDr: You may consider the newly added Assertments that ALu added.
This seems like a somewhat copy of functionality?

SE-FDr (changes requested):

monticore-grammar/src/main/java/de/monticore/statements/mcvardeclarationstatements/_cocos/VarDeclarationNameAlreadyDefinedInScope.java
TomBursch: My bad used https://github.com/MontiCore/monticore/blob/dev/monticore-grammar/src/main/java/de/monticore/statements/mcvardeclarationstatements/_cocos/VarDeclarationInitializationHasCorrectType.java as a base

monticore-grammar/src/test/java/de/monticore/statements/cocos/VarDeclarationNameAlreadyDefinedInScopeTest.java
TomBursch: I don't think there is a good corresponding method. In most tests I want to check the exact count of errors, not that at least one of that type has matched. 

Maybe that would be a good addition @luepges? `assertHasFinding` method variant that stops after the first match?

luepges: `assertHasFinding` should only find & remove the first match,
`assertHasFindings` matches & removes all findings (the naming might be somewhat nonoptimal)

SE-FDr: In this case I would use
`assertHasFindingStartingWith` or rather `assertHasFindingsStartingWith`;
* The current version is very verbose
* You are recreating mostly the same logic
* The actual amount does not matter, as it currently has more error messages than we would like anyways
* Whether this is an error or not seems negligible. If you think differently, this most likely should be a new function in MCAssertments

PR #165, opened 2025-07-04
Fix escaping of quotation symbols ' and " in implicit tokens.

DennisPruene:
This fixes two similar bugs concerning single and double quotes in implicit tokens respectively.

The first bug occurs, when a single quote `'` is used as the last character in an implicit token.
Example:
```
ExampleProd = "d'";
```

Here, the generated Antlr token will be `'d'`, thus missing the single quote at the end. The expected behavior is that the single quote is escaped correctly, so in the example we expect `'d\''`.

The second bug occurs, whenever a double quote `"` is used in an implicit token.
Example:
```
ExampleProd2 = "d\"";
```

Here, the generated Antlr token will be `'d\"'`. This causes Antlr to throw an error, as escaping double quotes (`\"`) is not allowed in single quoted string literals. The expected behavior is that each occurrence of `\"` is replaced by `"` so in this example we expect the result `'d"'`.

### DIFF
--- a/monticore-generator/src/main/java/de/monticore/codegen/parser/ParserGeneratorHelper.java
+++ b/monticore-generator/src/main/java/de/monticore/codegen/parser/ParserGeneratorHelper.java
@@ -159,15 +159,22 @@ public class ParserGeneratorHelper {
     }
   }
 
+  /**
+   * @param str - A String whose contents were taken directly from a StringLiteral.
+   * @return The original string, but each occurrence of ' is replaced with \'.
+   */
+  @SuppressWarnings("unused") // Used in parser/Lexer.ftl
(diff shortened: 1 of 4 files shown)


luepges:
Could you include a test grammar (e.g., in the monticore-test/02.experiments/parser) subproject?

(The MontiVerse pipeline will fail due to GitHub's token scope, I'll trigger it by hand)

DennisPruene:
I've added a test grammar along with a few tests.

luepges:
[#1756004](https://git.rwth-aachen.de/monticore/montiverseci/-/pipelines/1756004)

SchmidtMarc (approved):

PR #167, opened 2025-07-09
Fix: Builder for Terminals present in the AST

luepges:
Fix that, e.g. LogicalAndExpressions require a `setOperator("&&")` call when using their builders.

Also fix stringliteral decoding of single character length, such as `"/"`


SE-FDr (approved):
There could be some open cases, I am not sure.-
But if, those could be solved in a follow-up.

PR #168, opened 2025-07-10
Fix/terminal attribute order

luepges:
In addition to #167 

Interfaces with ast-present terminals (such as `InfixExpressions`) result in multiple attributes within the CD.
The last attribute used to be kept, resulting in `BooleanAndOpExpression implements Expression <120>, InfixExpression =
    left:Expression operator:"&&" right:Expression;` having the `operator` default being `""` (from the `InfixExpression`)

This PR modifies the removal of redundant attributes, by
 * cleaning up the code
 * preferring the attribute with the highest category (i.e., List), then the highest count of stereotypes (i.e., `<<inherited>>`), and finally prefers the first attribute (which is the original one) over later ones (e.g., from the infixexpression interface)

### DIFF
--- a/monticore-generator/src/main/java/de/monticore/codegen/mc2cd/manipul/RemoveRedundantAttributesManipulation.java
+++ b/monticore-generator/src/main/java/de/monticore/codegen/mc2cd/manipul/RemoveRedundantAttributesManipulation.java
@@ -12,8 +12,8 @@ import de.monticore.codegen.mc2cd.TransformationHelper;
 import de.monticore.types.mccollectiontypes._ast.ASTMCGenericType;
 import de.monticore.types.mccollectiontypes._ast.ASTMCTypeArgument;
 
+import java.util.Iterator;
 import java.util.List;
-import java.util.ListIterator;
 import java.util.Optional;
 import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


monticore-generator/src/main/java/de/monticore/codegen/mc2cd/manipul/RemoveRedundantAttributesManipulation.java
```diff
@@ -42,44 +41,74 @@ public ASTCDCompilationUnit apply(ASTCDCompilationUnit cdCompilationUnit) {
...
+    } else if (categoryRelation > 0) {
+      return true;
+    }
+    // The same category: check stereo counts to keep inherited
+    int candStereoCount = candAttr.getModifier().isPresentStereotype()
+            ? candAttr.getModifier().getStereotype().getValuesList().size() : 0;
+    int existingStereoCount = existingAttr.getModifier().isPresentStereotype()
+            ? existingAttr.getModifier().getStereotype().getValuesList().size() : 0;
+    return candStereoCount > existingStereoCount;
+    // By default, the first attribute is kept
```
SE-FDr: * Mark as heuristic?
* Log this decision (at least in the case of the less likely case), as this is not behavior that is easy to anticipate, and otherwise very hard to find/confusing.

SE-FDr (changes requested):

SE-FDr (approved):

PR #175, opened 2025-07-24
Add MCJavaLiteralsInterpreter

jra-se:
The old interpreter needs an implementation for the MCJavaLiteralsInterpreter as Java-based languages use MCJavaLiterals instead of MCCommonLiterals.


SE-FDr (approved):
We need tests. Future issue

PR #182, opened 2025-08-07
Fix project#getGroup() not being respected

luepges:
Project#getGroup() is only available after the project has been evaluated.


luepges:
See https://github.com/MontiCore/cd4analysis/pull/67

ahellwig (approved):

PR #184, opened 2025-08-11
Prevent empty astrules in generated DSTLs

jra-se:
A generated DSTL contains ASTRules mirroring the original grammar.
However, these ASTRules do NOT contain methods and also attributs only under certain conditions.
In many cases, this results in empty ASTRules in the DSTL, that have no impact and could be removed.

**Example:**
```
astrule ITFODName;
```

This pull requests adds checks for necessity befor adding an ASTRule to the DSTL.


luepges:
Please include a test grammar which would cause problems without this fix
 (e.g., in [monticore-test/montitrans/test-dstl-gen](https://github.com/MontiCore/monticore/tree/dev/monticore-test/montitrans/test-dstl-gen/src/main/grammars/mc/testcases))

luepges (approved):

PR #185, opened 2025-08-20
Fix override annotations for generated dstl prods

jra-se:
This Pull Request adds `@Override` annotations for generated DSTLs in case of the generation of the typical TF production rules for a non-terminal that is overridden (4762).

**Example**
```
grammar X {
  A = Name;
}

grammar Y extends X {
  @Override
  A = Name | Digits;
}
```

The generated DSTLs will look like the following:
```
grammar XTF {
  interface ITFA;
  A_Neg implements ITFA ... = ...
  ...
}

grammar YTF extends XTF {
  A_Neg implements ITFA ... = ...
  ...
}
```

Due to the missing `@Override` annotation at the `A_Neg` non-terminal, the two generated `ASTA_Neg` (for grammar `XTF` and grammar `YTF`) are independent. Instead, the `ASTA_Neg` of `YTF` should extend `ASTA_Neg` of `XTF`.
```
grammar YTF extends XTF {
  @Override
  A_Neg implements ITFA ... = ...
  ...
}
```

### DIFF
--- a/monticore-generator/src/main/java/de/monticore/dstlgen/grammartransformation/DSL2TransformationLanguageVisitor.java
+++ b/monticore-generator/src/main/java/de/monticore/dstlgen/grammartransformation/DSL2TransformationLanguageVisitor.java
@@ -357,6 +357,7 @@ public class DSL2TransformationLanguageVisitor implements
   }
   
   protected void addASTClassProdAnnotations(ASTClassProd prod, boolean override) {
+    // Add @Override annotation to the production if needed
     if (override) {
       prod.addGrammarAnnotation(productionFactory.createOverrideAnnotation());
     }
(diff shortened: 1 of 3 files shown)


monticore-generator/src/main/java/de/monticore/dstlgen/grammartransformation/DSL2TransformationLanguageVisitor.java
```diff
@@ -350,5 +335,31 @@ public void visit(ASTConcept srcNode) {
...
+        return false;
+      }
+      productionFactory.addInterfaces(srcNode.getSuperRuleList(), prod);
+      productionFactory.addInterfaces(srcNode.getSuperInterfaceRuleList(), prod);
+    }
+    return true;
+  }
+  
+  protected void addASTClassProdAnnotations(ASTClassProd prod, boolean override) {
+    if (override) {
```
luepges, line 361:
```suggestion
    // Add @Override annotation to the production if needed
    if (override) {
```

(This class lacks documentation/comments, so why not start now :) )

luepges (approved):

luepges:
> Due to the missing @Override annotation at the A_Neg non-terminal, the two generated ASTA_Neg [...] are independent
Afaik, the `@Override` annotation is optional and only a warning will be logged (which I have sucessfully ignored the last few years).
The `ytf.ASTA_Neg` should implicitly extend/override the `xtf.ASTA_Neg`

PR #186, opened 2025-08-26
Add missing replacement of free type vars in resolveConstructors

jra-se:


SE-FDr (approved):
Tests already exists.

PR #198, opened 2025-10-22
don't normalize before function split

SE-FDr:
Did not test offline -> check pipeline status before merge


luepges:
small test for the erroring scenario👼

SE-FDr:
small test for the erroring scenario👼, indeed.

luepges (approved):

PR #200, opened 2025-10-30
#4835 improve error handling guava

Hendrik7889:
Completed subtasks:
    Log::errorIfNull should be deprecated and replaced   
    Objects::requireNonNull should be replaced as well (technically not part of the Logging issue, but related)

There are still open subtasks in in issue #4835


monticore-generator/src/main/java/de/monticore/DelegatingClassLoader.java
```diff
@@ -1,6 +1,9 @@
 /* (c) https://github.com/MontiCore/monticore */
 package de.monticore;
 
+import com.google.common.base.Preconditions;
+import com.google.common.base.Verify;
```
SE-FDr:
```suggestion
```

monticore-runtime/src/main/java/de/monticore/generating/templateengine/freemarker/FreeMarkerTemplateEngine.java
```diff
@@ -8,7 +8,9 @@
 import java.io.StringWriter;
 import java.io.Writer;
 import java.lang.reflect.InvocationTargetException;
+import static com.google.common.base.Preconditions.checkNotNull;
```
SE-FDr:
```suggestion
```

SE-FDr (changes requested):
Seems good.

SE-FDr (approved):
The FD project needs to be updated accordingly, as far as I have seen.
Afterwards, this should be mergeable.

PR #212, opened 2025-11-21
Fix symbol surrogate equals

schmalzing:
Symbols should be usable without knowing that surrogates exist. A surrogate therefore loads its delegate when needed to provide access to the members of the actual symbol. For `equals(Object)`, this was so far not the case. This merge request fixes this problem.

However, the fix is not straight forward. One may call `surrogate.equals(symbol)` or `symbol.equals(surrogate)` and in both cases the result should be true iff the delegate of the surrogate is the symbol. Furthermore, one has do consider that the symbol may be another surrogate. Therefore, `equals` of the symbol has to be adapted as well. 

The solution requires the use of `instance of`, which we would have like to avoid. However, its the most lightweight solution we found. Furthermore, equals is the one case where knowledge of surrogates affects a non-surrogate implementation.

### DIFF
--- a/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolSurrogateDecorator.java
+++ b/monticore-generator/src/main/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolSurrogateDecorator.java
@@ -32,6 +32,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import static de.monticore.cd.codegen.CD2JavaTemplates.ANNOTATIONS;
 import static de.monticore.cd.codegen.CD2JavaTemplates.EMPTY_BODY;
 import static de.monticore.cd.facade.CDModifier.PROTECTED;
 import static de.monticore.cd.facade.CDModifier.PUBLIC;
@@ -125,6 +126,7 @@ public class SymbolSurrogateDecorator extends AbstractCreator<ASTCDClass, ASTCDC
       .addAllCDMembers(nameMethods)
       .addAllCDMembers(delegateSymbolRuleAttributeMethods)
       .addAllCDMembers(delegateAccecptMethods)
+      .addCDMember(createGetThis(this.symbolTableService.getSymbolSimpleName(symbolInput)))
       .addCDMember(createGetFullNameMethod())
       .addCDMember(createOverridenDeterminePackageName())
       .addCDMember(createOverridenDetermineFullName())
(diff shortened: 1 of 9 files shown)


monticore-runtime/src/main/java/de/monticore/symboltable/ISymbol.java
```diff
@@ -96,4 +97,7 @@ default void accept (ITraverser visitor)  {
     visitor.handle(this);
   }
 
+  default ISymbol getThis() {
```
SE-FDr: do we want this on the public level?
If we add it in each generated symbol,
we can make it protected, as we currently (and hopefully fully)
only ever need it for `equals()` (and perhaps hashvalue).

monticore-generator/src/main/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolSurrogateDecorator.java
```diff
@@ -145,13 +147,21 @@ protected ASTCDMethod createSetEnclosingScopeMethod(ASTCDAttribute enclosingScop
     this.replaceTemplate(EMPTY_BODY, method, new TemplateHookPoint(TEMPLATE_PATH + "SetEnclosingScope4SymbolSurrogate", enclosingScopeAttribute, scopeName));
     return method;
   }
+
+  protected ASTCDMethod createGetThis(String symbolClass) {
+    ASTCDMethod method = getCDMethodFacade().createMethod(PUBLIC.build(), symbolClass, "getThis");
```
SE-FDr: s. below. Any use case for making this public?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/_symboltable/symbol/SymbolSurrogateDecorator.java
```diff
@@ -145,13 +147,21 @@ protected ASTCDMethod createSetEnclosingScopeMethod(ASTCDAttribute enclosingScop
...
+    return method;
+  }
   
   protected ASTCDConstructor createConstructor(String symbolSurrogateClass) {
     ASTCDParameter nameParameter = getCDParameterFacade().createParameter(String.class, NAME_VAR);
     ASTCDConstructor constructor = getCDConstructorFacade().createConstructor(PUBLIC.build(), symbolSurrogateClass, nameParameter);
     this.replaceTemplate(EMPTY_BODY, constructor, new TemplateHookPoint(TEMPLATE_PATH + "ConstructorSymbolSurrogate"));
     return constructor;
   }
+
```
SE-FDr, line 164:
```suggestion
```

schmalzing:
@rumpe for information

monticore-runtime/src/main/java/de/monticore/symboltable/ISymbol.java
schmalzing: Wird jetzt generiert (protected)

SE-FDr (approved):

PR #214, opened 2025-11-26
Fix non-determinism, remove hashset

schmalzing:
MontiCore's resolve mechanisms should behave deterministically. That is, return the same symbols in the same order on the same resolve calls. However, `getResolvedOrThrowException` uses a HashSet, potentially to remove duplicates, which results in a collection of symbols in arbitrary order. 

Removing duplicates is neither the task of that method nor is it needed. Methods that call this method already remove duplicates. Furthermore, the symbol table should be considered broken if local symbol lists contain duplicates.


SE-FDr (approved):
Assuming the Pipeline runs through.

PR #227, opened 2025-12-14
Fix equals between incompatible types

schmalzing:


luepges (changes requested):
+test um regression zu verhinden

€: We are not quite sure what exactly causes the problem:
```
else if ("de.monticore.symbols.basicsymbols._symboltable.FunctionSymbol".equals(kind.get())
      || "de.monticore.symbols.basicsymbols._symboltable.FunctionSymbol".equals(deSer.getSerializedKind())) {
    de.monticore.symbols.basicsymbols._symboltable.FunctionSymbol s1 = (de.monticore.symbols.basicsymbols._symboltable.FunctionSymbol) deSer.deserialize(scope, symbol);
    scope.add(s1);
    scope.addSubScope(s1.getSpannedScope());
  }
```
At some point `deSer.getSerializedKind()` is not equals to `kind.get() `, but when?

luepges:
Created seperate issue for the regression test: monticore/monticore/-/issues/4934

luepges (approved):

PR #229, opened 2025-12-16
Move the Gradle Entrypoint to the generator subproject 

luepges:
…(as the gradle-plugin should not be a part of the mcTool classpath)


monticore-generator/gradle-plugin/src/test/java/de/monticore/gradle/MCGenPluginTest.java
```diff
@@ -132,12 +132,14 @@ void testGenerateGrammar(String version) throws IOException {
     writeFile(new File(grammarDir, "MyTestGrammarS.mc4"),
             "grammar MyTestGrammarS extends MyTestGrammar { Monti = \"Core\"; }");
 
+    // use a custom gradle home directory to ensure fresh cashes
```
luepges:
```suggestion
    // use a custom gradle home directory to ensure fresh caches
```

luepges:
MontiVerse ran into Gateway Time-out after typo commit, passes other checks. Check by CBu

PR #233, opened 2026-01-04
Typecheck fixes for OCL flattening

ahellwig:


SE-FDr:
Does not work in every case,
PR replaced with https://github.com/MontiCore/monticore/pull/234

PR #235, opened 2026-01-09
Fix/gitattributes

luepges:
Override git auto-text dedection to ensure jars & images are always binary files (and not text files, as it sometimes assumes)


.gitattributes
```diff
@@ -1,3 +1,5 @@
 # All files (*) that git considers to be text (auto)
 # will have normalized (LF) line-endings in the repository.
-*  text=auto eol=lf
+*      text=auto eol=lf
```
SE-FDr, line 3:
```suggestion
* text=auto eol=lf
```

SE-FDr (approved):

.gitattributes
luepges: In the git manpages, the file appears to be spaced using columns (https://git-scm.com/docs/gitattributes)

luepges:
We might have to add other file extensions, in case they cause problems too.

PR #238, opened 2026-01-22
delegate equality workaround

SE-FDr:
quick fix for GuiDSL, needs proper solution


luepges (approved):
Unhappy, aber sinnvoll

PR #243, opened 2026-02-06
Deprecate setDim method in SymTypeArray

SE-FDr:


luepges (approved):
(assuming pipeline passes - we could skip the MontiVerse for this one)

PR #245, opened 2026-02-10
Endless resolution recursionv – fix

SE-FDr:


schmalzing (changes requested):
Add regression tests to prevent reintroducing degeneration issues, such as a test case that previously triggered infinite recursion but now terminates successfully.

SE-FDr:
converted to draft to extend the implementation to fix another issue first, as that allows for easier test implementation.

SE-FDr:
I don't like it, but I cannot find the case there this can happen.
(It might be that I found this case due to another bug somewhere else that has been fixed already.,)
However, I would still merge it simply as "more performant and cleaner solution".

luepges (approved):
(unable to find regression test-case)

PR #250, opened 2026-02-25
Javagen: Fix casts in equals operator printing; Convenience methods for printing

ahellwig:


SE-FDr (approved):
From my side this seems fine, but I edited a lot.
@ahellwig you may check it again before merging.

PR #269, opened 2026-04-13
Lower spotbugs version to avoid junit dependency bug

ahellwig:
Otherwise monticore-runtime has a transitive api dependency on junit-bom, which forces downstream projects to a specific JUnit version

See https://github.com/spotbugs/spotbugs/issues/3908


luepges:
The fix is still unreleased 
> Unreleased - 2025-??-??
https://github.com/spotbugs/spotbugs/blob/ca621d4096b099ef2b086715320468cc8c1259ac/CHANGELOG.md

luepges (approved):

PR #270, opened 2026-04-15
Deprecate Surrogates

SE-FDr:
See #273 for the suggested replacement


MisterErwin (approved):
FYI: we are replacing surrogates with suppliers due to incorrect inheritance of the surrogate class. (BasicTypeSymbolSurrogate does not extend OOTypeSymbol, but the lazily fetched symbol might be a OOTyeSymbol)

luepges (approved):

PR #281, opened 2026-05-05
Fix ExpressionJavaGenerator

TomBursch:
- String-based Java type printing
- Add stream type printing support
- Update some conversion handler rules to avoid overlap
- Fix some minor bugs with static variables


monticore-grammar/src/main/java/de/monticore/codegen/javagen/operationprinter/JavaNumericOperationHandler.java
```diff
@@ -132,18 +131,17 @@ protected void printSimpleInfixOperator(
...
     return isNumericType(modelType)
         && (
-        isByte(javaType)
-            || isShort(javaType)
-            || isChar(javaType)
-            || isInt(javaType)
-            || isLong(javaType)
-            || isFloat(javaType)
-            || isDouble(javaType)
+        isByte(modelType)
```
SE-FDr: This does not work.
SIUnits are Numeric Types in Java
-> we will need a check
`isNumericTypeInJava` or something like it.

monticore-grammar/src/main/java/de/monticore/codegen/javagen/operationprinter/JavaEqualityOperationHandler.java
```diff
@@ -143,16 +139,16 @@ protected void printRightConverted(
...
-    return javaType.isPrimitive() && (
-        isBoolean(javaType)
-            || isByte(javaType)
-            || isShort(javaType)
-            || isChar(javaType)
-            || isInt(javaType)
-            || isLong(javaType)
-            || isFloat(javaType)
-            || isDouble(javaType)
+  protected boolean isPrimitive(SymTypeExpression type) {
```
SE-FDr: Same as below: We need `isPrimitiveInJava`

monticore-grammar/src/main/java/de/monticore/codegen/javagen/typeconverter/JavaObjectConversionHandler.java
```diff
@@ -22,10 +23,13 @@ public boolean tryPrintConverted(
       SymTypeExpression modelSourceType,
       CodeGenPrintAction sourceExprPrintAction
   ) {
-    if ((modelTargetType.isObjectType() || modelTargetType.isGenericType()) &&
-        (modelSourceType.isObjectType() || modelSourceType.isGenericType())
+    if (((modelTargetType.isObjectType() || modelTargetType.isGenericType()) && (modelSourceType.isObjectType() || modelSourceType.isGenericType()) ||
```
SE-FDr: Nobody can read it, split?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/typeconverter/JavaObjectToStringConversionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+      CodeGenPrintAction sourceExprPrintAction
+  ) {
+    if (isStringOrSubType(modelTargetType) && (modelSourceType.isObjectType() || modelSourceType.isGenericType()) &&
+        !(modelTargetType.isObjectType() || modelTargetType.isGenericType()) && (modelSourceType.isObjectType() || modelSourceType.isGenericType()) &&
+        !((modelTargetType.isObjectType() || modelTargetType.isGenericType()) && isStringOrSubType(modelSourceType)) &&
+        !(isStringOrSubType(modelTargetType) && isStringOrSubType(modelSourceType))
+    ) {
+      printer.print("(");
+      sourceExprPrintAction.print(printer);
+      printer.print(").toString()");
```
SE-FDr: I think this does not work for some TypeVariables?
It should probably be `String.valueOf()` or something along that.

monticore-grammar/src/main/java/de/monticore/codegen/javagen/typeconverter/JavaToStringConversionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @Override
+  public boolean tryPrintConverted(
+      IndentPrinter printer,
+      SymTypeExpression modelTargetType,
+      SymTypeExpression modelSourceType,
+      CodeGenPrintAction sourceExprPrintAction
+  ) {
+    if (isStringOrSubType(modelTargetType) && (modelSourceType.isObjectType() || modelSourceType.isGenericType()) &&
+        !(modelTargetType.isObjectType() || modelTargetType.isGenericType()) && (modelSourceType.isObjectType() || modelSourceType.isGenericType()) &&
```
SE-FDr, line 26: I think this needs a comment as well or be split, I have a hard time following the logic

monticore-grammar/src/main/java/de/monticore/codegen/javagen/typeconverter/JavaTupleConversionHandler.java
```diff
@@ -35,15 +34,15 @@ public boolean tryPrintConverted(
...
       printer.print(">) () -> { ");
-      printer.print(printJavaType(getAsJavaType(sourceTuple)));
+      printer.print(printJavaType(sourceTuple));
       printer.print(" ");
       printer.print(tmpTupleVarName);
       printer.print(" = ");
       sourceExprPrintAction.print(printer);
       printer.print("; return ");
-      printer.print(getAsJavaType(targetTuple).asGenericType().getTypeConstructorFullName());
+      printer.print(printJavaType(targetTuple).split("<")[0]);
```
SE-FDr: That should not be part of this specifc  printer, we should not work on strings (and if we do, hide it)
-> `printJavaTypeConstructorFullName`?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/SymTypeExpressionBoxedJavaPrinterVisitor.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Prints SymTypeExpressions in a Java compatible way,
+ */
+public class JavaBoxedSymTypeExpressionPrinterVisitor
+    extends JavaSymTypeExpressionPrinterVisitor {
+
+  protected Map<String, String> boxMap;
```
SE-FDr, line 18: static?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/SymTypeExpressionBoxedJavaPrinterVisitor.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  public void visit(SymTypePrimitive symType) {
+    getPrint().append(box(symType.getPrimitiveName()));
+  }
+
+  @Override
+  public void visit(SymTypeOfSIUnit siUnit) {
+    // to be extended
+    getPrint().append(box("double"));
```
SE-FDr: Why is code here if it is not complete?
Remove or complete it?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/SymTypeExpressionBoxedJavaPrinterVisitor.java
```diff
@@ -0,0 +1,48 @@
...
+    Log.error("0xFD240 conversion of SIUnit to Java is not supported yet.");
+  }
+
+  public String box(String unboxedName) {
+    return boxMap.getOrDefault(unboxedName, unboxedName);
```
SE-FDr, line 44: does not sound like something that should be public

monticore-grammar/src/main/java/de/monticore/codegen/javagen/SymTypeExpressionBoxedJavaPrinterVisitor.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Prints SymTypeExpressions in a Java compatible way,
+ */
+public class JavaBoxedSymTypeExpressionPrinterVisitor
+    extends JavaSymTypeExpressionPrinterVisitor {
+
+  protected Map<String, String> boxMap;
+
+  public JavaBoxedSymTypeExpressionPrinterVisitor() {
+    Map<String, String> boxMap_temp = new LinkedHashMap<>();
```
SE-FDr: Additionally, Collection types are missing

SE-FDr: And String

monticore-grammar/src/main/java/de/monticore/codegen/javagen/typeconverter/JavaObjectToStringConversionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+import de.monticore.codegen.CodeGenPrintAction;
+import de.monticore.prettyprint.IndentPrinter;
+import de.monticore.types.check.SymTypeExpression;
+
+import static de.monticore.types3.SymTypeRelations.isStringOrSubType;
+
+/**
+ * Conversions of Objects (including generics) to string
+ */
+public class JavaObjectToStringConversionHandler
```
SE-FDr: Why do we need a JavaObjectToString Conversion handler,
why not a ToStringConversion hander? What are the cases?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/SymTypeExpressionBoxedJavaPrinterVisitor.java
SE-FDr: Ah, wait, you have split it into multiple visitors, kay...
In that case, the SymTypeOfSIUnit is odd here, that is not specific to this visitor.

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaSymTypeExpressionPrinterVisitor.java
```diff
@@ -1,68 +1,186 @@
...
  */
 public class JavaSymTypeExpressionPrinterVisitor
     extends SymTypePrintFullNameVisitor {
 
-  @Override
-  public void visit(SymTypeInferenceVariable infVar) {
-    logKindIsUnsupported(infVar);
+  protected final String RTE_PACKAGE = "de.monticore.rte";
+
+  protected Map<String, String> overrides;
```
SE-FDr: same, static

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaSymTypeExpressionPrinterVisitor.java
```diff
@@ -1,68 +1,186 @@
...
- * {@link SymTypeExpression2JavaConverter}.
+ * Prints SymTypeExpressions in a Java compatible way,
  */
 public class JavaSymTypeExpressionPrinterVisitor
     extends SymTypePrintFullNameVisitor {
 
-  @Override
-  public void visit(SymTypeInferenceVariable infVar) {
-    logKindIsUnsupported(infVar);
+  protected final String RTE_PACKAGE = "de.monticore.rte";
```
SE-FDr: should not be in multiple positions, but I think this is the only one? Then it should be fine here

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaSymTypeExpressionPrinterVisitor.java
SE-FDr: Also, the name `overrides` seems strictly wrong.

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaSymTypeExpressionPrinterVisitor.java
```diff
@@ -1,68 +1,186 @@
...
+  public void visit(SymTypeOfFunction func) {
+    // precondition; support could be extended if required
+    if (func.isElliptic()) {
+      Log.error("0xFD324 internal error:" +
+          "No support for elliptic Functions exists yet.");
+      return;
+    }
+
+    // Main Symbol: Function or Action?
+    String className;
```
SE-FDr: We are doing to much with Strings here -> maybe one StringBuilder in the class that is used by every method?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaSymTypeExpressionPrinterVisitor.java
```diff
@@ -1,68 +1,186 @@
...
+    if (isFunc) {
+      resArgs.add(func.getType());
+    }
+    resArgs.addAll(func.getArgumentTypeList());
+
+    if (!resArgs.isEmpty()) {
+      getPrint().append('<');
+      for (int i = 0; i < resArgs.size(); i++) {
+        SymTypeExpression innerType = resArgs.get(i);
+        printOpeningBracketForInner(innerType);
```
SE-FDr: what? why? what is this?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/SymTypeExpressionJavaPrinterVisitor.java
```diff
@@ -1,68 +1,186 @@
...
+      for (int i = 0; i < resArgs.size(); i++) {
+        SymTypeExpression innerType = resArgs.get(i);
+        printOpeningBracketForInner(innerType);
+        getPrint().append(SymTypeExpression2JavaConverter.printBoxedJavaType(innerType));
+        printClosingBracketForInner(innerType);
+        if (i < resArgs.size() - 1) {
+          getPrint().append(',');
+        }
+      }
+      getPrint().append('>');
```
SE-FDr, line 84: `getPrint()` is odd, getPrinter?

SE-FDr: Not entirely incorrect, but a bit odd.

SE-FDr: also does not fit the other methods in this module which are more getPrinter

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaSymTypeExpressionPrinterVisitor.java
SE-FDr: yes, the other parts use getPrint(). why not that?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaSymTypeExpressionPrinterVisitor.java
```diff
@@ -1,68 +1,186 @@
...
+        getPrint().append(SymTypeExpression2JavaConverter.printBoxedJavaType(innerType));
```
SE-FDr: ah, because it is boxed, fine

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaSymTypeExpressionPrinterVisitor.java
```diff
@@ -1,68 +1,186 @@
...
   public void visit(SymTypeOfNull nullType) {
-    logKindIsUnsupported(nullType);
+    Log.error("0xFD322 internal error: "
+        + "Tried to convert SymTypeOfNull to Java");
   }
 
   @Override
   public void visit(SymTypeOfRegEx regEx) {
-    logKindIsUnsupported(regEx);
+    getPrint().append("java.lang.String");
```
SE-FDr: We use this multiple times -> maybe we need constants for the printer

monticore-grammar/src/main/java/de/monticore/codegen/javagen/SymTypeExpression2JavaConverter.java
```diff
@@ -2,107 +2,42 @@
...
 
-  protected String _printJavaType(SymTypeExpression javaType) {
-    return javaTypePrinterVisitor.calculate(javaType);
+  protected String _printJavaType(SymTypeExpression modelType) {
+    return javaTypePrinterVisitor.calculate(modelType);
   }
 
   /**
-   * applies type erasure to the given Java compatible type
+   * Prints Java compatible types in a Java compatible way.
```
SE-FDr: Where did the type erasure go to? this method seems missing now?

monticore-grammar/src/main/java/de/monticore/expressions/assignmentexpressions/codegen/javagen/AssignmentExpressionsJavaGenVisitor.java
```diff
@@ -61,12 +62,34 @@ public AssignmentExpressionsJavaGenVisitor(IndentPrinter printer) {
...
+    SymTypeExpression innerType = normalize(typeOf(expr.getExpression()));
+
+    JavaOperationPrinter.printAssignment(
+        printer,
+        resultType,
+        innerType,
+        // left type due to conversion
+        innerType,
+        p -> expr.getExpression().accept(getTraverser()),
+        p2 -> printPlus(printer, resultType, innerType, SymTypeExpressionFactory.createPrimitive("int"), p -> expr.getExpression().accept(getTraverser()), (p) -> p.print("1"))
```
SE-FDr, line 109: Am I reading this wrong? this does not seem to do what it should do, right?

monticore-grammar/src/main/java/de/monticore/expressions/commonexpressions/codegen/javagen/CommonExpressionsJavaGenVisitor.java
```diff
@@ -416,17 +415,28 @@ public void handle(ASTFieldAccessExpression node) {
...
 
     Type4Ast type4Ast = _internal_hacky_trafo_workaround_dont_use_overwise();
 
+    // static fields
+    if (type4Ast.hasPartialTypeOfTypeIdentifierForName(node.getExpression())) {
+      SymTypeExpression innerType = type4Ast.getPartialTypeOfTypeIdForName(node.getExpression());
+      getPrinter().print(printJavaType(innerType));
+      getPrinter().print(".");
+      getPrinter().print(node.getName());
+      return;
```
SE-FDr: We may want to avoid multiple return points in a method

SE-FDr: s. BR

monticore-grammar/src/main/java/de/monticore/expressions/commonexpressions/codegen/javagen/CommonExpressionsJavaGenVisitor.java
```diff
@@ -416,17 +415,28 @@ public void handle(ASTFieldAccessExpression node) {
 
     // NOTE: can be (should be) extended to provide type parameters
 
-    SymTypeExpression outerType = normalize(typeOf(node.getExpression()));
 
     Type4Ast type4Ast = _internal_hacky_trafo_workaround_dont_use_overwise();
 
+    // static fields
+    if (type4Ast.hasPartialTypeOfTypeIdentifierForName(node.getExpression())) {
+      SymTypeExpression innerType = type4Ast.getPartialTypeOfTypeIdForName(node.getExpression());
```
SE-FDr: Does this work in all cases? I am currently unsure.

SE-FDr: It probably does

monticore-grammar/src/main/java/de/monticore/expressions/commonexpressions/codegen/javagen/CommonExpressionsJavaGenVisitor.java
```diff
@@ -469,17 +479,10 @@ else if (methodSym.isIsStatic()) {
...
-    else if (type4Ast.hasPartialTypeOfTypeIdentifierForName(node.getExpression())) {
-      SymTypeExpression innerType = type4Ast.getPartialTypeOfTypeIdForName(node.getExpression());
-      getPrinter().print(printJavaType(getAsJavaType(innerType)));
-      getPrinter().print(".");
-      getPrinter().print(node.getName());
-    }
     // non-static fields
     else if (
         outerType.getSourceInfo().getSourceSymbol()
-            .filter(s -> s instanceof FieldSymbol)
```
SE-FDr, line 483: Why?

monticore-grammar/src/main/java/de/monticore/expressions/expressionsbasis/codegen/javagen/ExpressionsBasisJavaGenVisitor.java
```diff
@@ -51,15 +53,32 @@ public void handle(ASTNameExpression node) {
 
     // NOTE: can be (should be) extended to provide type parameters
 
+    Type4Ast type4Ast = _internal_hacky_trafo_workaround_dont_use_overwise();
+
+
+    // type reference
+    if (type4Ast.hasPartialTypeOfTypeIdentifierForName(node)) {
```
SE-FDr: Which case does this solve? This should never be triggered, or am I missing a case?

monticore-grammar/src/main/java/de/monticore/expressions/expressionsbasis/codegen/javagen/ExpressionsBasisJavaGenVisitor.java
```diff
@@ -51,15 +53,32 @@ public void handle(ASTNameExpression node) {
...
+    }
+
+
     SymTypeExpression exprType = normalize(typeOf(node));
 
+    if (exprType.getSourceInfo().getSourceSymbol().isPresent()
+        && OOSymbolsMill.typeDispatcher().isOOSymbolsField(exprType.getSourceInfo().getSourceSymbol().get())
+        && OOSymbolsMill.typeDispatcher().asOOSymbolsField(exprType.getSourceInfo().getSourceSymbol().get()).isIsStatic()) {
+      getPrinter().print(OOSymbolsMill.typeDispatcher().asOOSymbolsField(exprType.getSourceInfo().getSourceSymbol().get()).getFullName());
+    }
```
SE-FDr, line 63: what happens here, what case is this?

monticore-grammar/src/main/java/de/monticore/expressions/tupleexpressions/codegen/javagen/TupleExpressionsJavaGenVisitor.java
```diff
@@ -7,13 +7,12 @@
...
 import de.monticore.prettyprint.IndentPrinter;
-import de.monticore.types.check.SymTypeExpression;
-import de.monticore.types.check.SymTypeOfGenerics;
+import de.monticore.types.check.SymTypeOfTuple;
 import de.monticore.types3.TypeCheck3;
 
 import java.util.Iterator;
 
-import static de.monticore.codegen.javagen.SymTypeExpression2JavaConverter.getAsJavaType;
+import static de.monticore.codegen.javagen.SymTypeExpression2JavaConverter.printBoxedJavaType;
```
SE-FDr: wait, now I know what was bothering me
-> `printAsBoxedJavaType`

SE-FDr: scratch that, it does not even print, it returns a String, so it should not be called `print`
-> `getAsBoxedJavaTypeString`?

SE-FDr: maybe there is a neater name

monticore-grammar/src/test/java/de/monticore/codegen/javagen/AbstractJavaGenTest.java
```diff
@@ -60,6 +64,15 @@ void beforeEach() {
     jshell = JShell.create();
   }
 
+  void addMaxValueToInteger() {
```
SE-FDr: protected?

SE-FDr (changes requested):

monticore-grammar/src/main/java/de/monticore/expressions/commonexpressions/codegen/javagen/CommonExpressionsJavaGenVisitor.java
TomBursch: Otherwise variables won't work, and the variable and field case is identical

monticore-grammar/src/main/java/de/monticore/expressions/assignmentexpressions/codegen/javagen/AssignmentExpressionsJavaGenVisitor.java
TomBursch: Why not? It generates `i++` to `i = i + 1` with the right side always being an integer, which then gets converted to the correct type.

monticore-grammar/src/main/java/de/monticore/expressions/expressionsbasis/codegen/javagen/ExpressionsBasisJavaGenVisitor.java
TomBursch: Static field/enum

monticore-grammar/src/main/java/de/monticore/expressions/tupleexpressions/codegen/javagen/TupleExpressionsJavaGenVisitor.java
TomBursch: Yeah I know the name is not great. I don't like adding the return type to the name.
To make it backwards compatible the name should've been `printModelTypeAsJavaType` and `printModelTypeAsBoxedJavaType` which I'm not a huge fan either

monticore-grammar/src/main/java/de/monticore/expressions/assignmentexpressions/codegen/javagen/AssignmentExpressionsJavaGenVisitor.java
SE-FDr: Is that not `++i`?
`i = i + 1` should probably return the already incremented value if I am not mistaken.

monticore-grammar/src/main/java/de/monticore/expressions/commonexpressions/codegen/javagen/CommonExpressionsJavaGenVisitor.java
SE-FDr: In that case the issue is elsewhere,
Fields _are_ variables.
-> TypeDispatcher?

monticore-grammar/src/main/java/de/monticore/expressions/expressionsbasis/codegen/javagen/ExpressionsBasisJavaGenVisitor.java
SE-FDr: At least to check if something is a field or enum value, we have methods for, I think in OOSymbolsRelations or simmilar-ish.

monticore-grammar/src/main/java/de/monticore/expressions/tupleexpressions/codegen/javagen/TupleExpressionsJavaGenVisitor.java
SE-FDr: maybe `convert2BoxedJavaType` would be enough, since it does not state that prints anymore

monticore-grammar/src/main/java/de/monticore/expressions/expressionsbasis/codegen/javagen/ExpressionsBasisJavaGenVisitor.java
SE-FDr: For enums/fields I have something to offer:
https://monticore.github.io/monticore/monticore-grammar/javadoc/de/monticore/symbols/oosymbols/types3/OOSymbolsSymTypeRelations.html

TomBursch: It only has a [sourceIsEnumConstant](https://monticore.github.io/monticore/monticore-grammar/javadoc/de/monticore/symbols/oosymbols/types3/OOSymbolsSymTypeRelations.html#sourceIsEnumConstant(de.monticore.types.check.SymTypeExpression)) method, which works for enums. But there is nothing for (static) fields. The [getSourceFieldSymbolIfAvailable](https://monticore.github.io/monticore/monticore-grammar/javadoc/de/monticore/symbols/oosymbols/types3/OOSymbolsSymTypeRelations.html#getSourceFieldSymbolIfAvailable(de.monticore.types.check.SymTypeExpression)) method is protected.

SE-FDr: Correct, as described in the annotation, Resolver-specific checks are not handeled, as this is about the typing itself, not about their availability. (so the static check still needs the dispatcher)

monticore-grammar/src/main/java/de/monticore/expressions/assignmentexpressions/codegen/javagen/AssignmentExpressionsJavaGenVisitor.java
TomBursch: You're right; this wasn't correct, but I also couldn't come up with a good solution to this. (So I'm printing `++` now)

monticore-grammar/src/main/java/de/monticore/expressions/commonexpressions/codegen/javagen/CommonExpressionsJavaGenVisitor.java
TomBursch: But variables are not fields. And in MontiArc we use Variable for most things

monticore-grammar/src/main/java/de/monticore/expressions/expressionsbasis/codegen/javagen/ExpressionsBasisJavaGenVisitor.java
TomBursch: So can we leave this as is?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/SymTypeExpressionBoxedJavaPrinterVisitor.java
TomBursch: I'm not sure why it's odd here; this is just boxing primitives, hence we do not need anything else in the map

monticore-grammar/src/main/java/de/monticore/expressions/lambdaexpressions/codegen/javagen/LambdaExpressionsJavaGenVisitor.java
```diff
@@ -47,25 +46,24 @@ public void handle(ASTLambdaExpressionBody node) {
...
     for (int i = 0; i < node.getLambdaParameters().sizeLambdaParameters(); i++) {
       ASTLambdaParameter par = node.getLambdaParameters().getLambdaParameter(i);
       String parName = par.getName();
-      SymTypeExpression parType = javaFuncType.getArgument(i + 1);
+      SymTypeExpression parType = funcType.getArgumentType(i);
       if (i != 0) {
         getPrinter().print(", ");
       }
-      getPrinter().print(printJavaType(getAsJavaType(parType)));
+      getPrinter().print(convert2BoxedJavaType(parType));
```
SE-FDr, line 66: What was our last statement about this name? convert still seems wrong...

monticore-grammar/src/main/java/de/monticore/codegen/javagen/SymTypeExpression2JavaConverter.java
```diff
@@ -2,107 +2,64 @@
...
-  public static String printModelTypeAsJavaType(ASTExpression expression) {
-    return getDelegate()._printModelTypeAsJavaType(expression);
-  }
-
-  public String _printModelTypeAsJavaType(ASTExpression expression) {
-    return printModelTypeAsJavaType(TypeCheck3.typeOf(expression));
+  /**
+   * Converts model types in java compatible way and returns their type constructor
+   */
+  public static String getJavaTypeConstructor(SymTypeExpression modelType) {
```
SE-FDr: does not fit the other names. It also returns a String, not the actual type constructor. -> make consistent?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/typeconverter/JavaToStringConversionHandler.java
```diff
@@ -0,0 +1,40 @@
...
+    extends AbstractJavaTypeConverter {
+
+  @Override
+  public boolean tryPrintConverted(
+      IndentPrinter printer,
+      SymTypeExpression modelTargetType,
+      SymTypeExpression modelSourceType,
+      CodeGenPrintAction sourceExprPrintAction
+  ) {
+    if (// Target Type is String:
```
SE-FDr, line 23: In which cases is this one used?
usually, something cannot be converted to String directly,
the only exception that I can currently think of is `+` for String?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaGenSymTypeRelations.java
```diff
@@ -0,0 +1,87 @@
...
+    javaNumericTypes_temp.add("int");
+    javaNumericTypes_temp.add("java.lang.Integer");
+    javaNumericTypes_temp.add("long");
+    javaNumericTypes_temp.add("java.lang.Long");
+    javaNumericTypes_temp.add("short");
+    javaNumericTypes_temp.add("java.lang.Short");
+    javaNumericTypes = Collections.unmodifiableSet(javaNumericTypes_temp);
+  }
+
+  public static boolean isJavaPrimitive(SymTypeExpression type) {
```
SE-FDr, line 51: isPrintedAsJavaPrimitive?

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaGenSymTypeRelations.java
```diff
@@ -0,0 +1,87 @@
...
+    return getDelegate()._isJavaPrimitive(type);
+  }
+
+  public boolean _isJavaPrimitive(SymTypeExpression type) {
+    return javaPrimitiveTypes.contains(SymTypeExpression2JavaConverter.convert2JavaType(type));
+  }
+
+  public static boolean isJavaNumeric(SymTypeExpression type) {
```
SE-FDr, line 62: same

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaSymTypeRelations.java
```diff
@@ -0,0 +1,87 @@
...
+
+import com.google.common.base.Preconditions;
+import de.monticore.types.check.SymTypeExpression;
+import de.se_rwth.commons.logging.Log;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+public class JavaSymTypeRelations {
```
SE-FDr: Uff, hard one...
JavaGenSymTypeRelations is still rather short? Or something like it.
JavaSymTypeRelations is definitly confusing, that should not be the name

monticore-grammar/src/main/java/de/monticore/expressions/assignmentexpressions/codegen/javagen/AssignmentExpressionsJavaGenVisitor.java
```diff
@@ -53,30 +61,70 @@ public void setTraverser(AssignmentExpressionsTraverser traverser) {
...
-    _willBeRemoved_logUnimplemented(expr);
+    // NOTE: this is only a temporary implementation,
+    // as in the future, templates provided by the symbols
+    // are to be used instead.
+
+    if (JavaSymTypeRelations.isJavaNumeric(TypeCheck3.typeOf(expr.getExpression()))) {
+      expr.getExpression().accept(getTraverser());
+      getPrinter().print("--");
+    } else {
+      Log.error("0xFD249 Unhandled increment suffix operator "
```
SE-FDr: don't use the same error code twice.

monticore-grammar/src/main/java/de/monticore/expressions/assignmentexpressions/codegen/javagen/AssignmentExpressionsJavaGenVisitor.java
```diff
@@ -53,30 +61,70 @@ public void setTraverser(AssignmentExpressionsTraverser traverser) {
...
+    SymTypeExpression innerType = normalize(typeOf(expr.getExpression()));
+
+    JavaOperationPrinter.printAssignment(
+        printer,
+        resultType,
+        innerType,
+        // left type due to conversion
+        innerType,
+        p -> expr.getExpression().accept(getTraverser()),
+        p2 -> printPlus(printer, resultType, innerType, SymTypeExpressionFactory.createPrimitive("int"), p -> expr.getExpression().accept(getTraverser()), (p) -> p.print("1"))
```
SE-FDr, line 109: why does it use int here? what about floating points?

SE-FDr (changes requested):

monticore-grammar/src/main/java/de/monticore/expressions/lambdaexpressions/codegen/javagen/LambdaExpressionsJavaGenVisitor.java
TomBursch: > maybe `convert2BoxedJavaType` would be enough, since it does not state that prints anymore

https://github.com/MontiCore/monticore/pull/281#discussion_r3207934137

monticore-grammar/src/main/java/de/monticore/codegen/javagen/typeconverter/JavaToStringConversionHandler.java
TomBursch: Yeah, I realized this is something MontiArc-specific. We overwrote the type check to allow this, but this is not the default. I can move it into the MontiArc project, but I think other projects also overwrote that behavior, and this conversion handler will never be called without the overwrite, so we could also leave it in here

monticore-grammar/src/main/java/de/monticore/expressions/assignmentexpressions/codegen/javagen/AssignmentExpressionsJavaGenVisitor.java
TomBursch: The type conversion as part of the `printPlus` should automatically convert the source type of both sides to the target type of whatever that expression is.

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaGenSymTypeRelations.java
TomBursch: I've added a comment to make it more clear. I'm not sure I like the isPrintedAs prefix.

SE-FDr: Yes, the prefix could/should be improved, if you have better suggestions?
But I would definitly say that the name has to be precise (even without comments),
so I rather take an ugly name over a wrong/misinterpretable name.

monticore-grammar/src/main/java/de/monticore/expressions/lambdaexpressions/codegen/javagen/LambdaExpressionsJavaGenVisitor.java
SE-FDr: semi, it just states that one gets a type. What does that mean?
* SymTypeExpression?
* String?
* Class<?>?
Sure, it should not be called `print` anymore, but the name is not quite exact enough in my opinion.

monticore-grammar/src/main/java/de/monticore/codegen/javagen/typeconverter/JavaToStringConversionHandler.java
SE-FDr: If it is solely for `+`, I think it is fine to keep it in MC,  it does not matter if the printer can do more than the TC allows.
However, if it is only for `+`, I would put it directly into the `+` logic, to reduce the complexity of the type conversions. I would claim that the conversion to String is not an actual "type conversion" similar to others anyway, but rather an on the level of an implicit function call.

monticore-grammar/src/main/java/de/monticore/expressions/assignmentexpressions/codegen/javagen/AssignmentExpressionsJavaGenVisitor.java
SE-FDr: Yes, that may be the case, but that still makes the `int` here rather odd...
Ah, nevermind, I get it, it is the type of 1, fair enough, that works in any case.

monticore-grammar/src/main/java/de/monticore/expressions/lambdaexpressions/codegen/javagen/LambdaExpressionsJavaGenVisitor.java
SE-FDr: `getBoxedJavaTypePrint()`?
Still ugly, I want something better, but at least it states more what it does (print as the noun, not the verb, as it returns text)

SE-FDr:
superseded by https://github.com/MontiCore/monticore/pull/310

monticore-grammar/src/main/java/de/monticore/expressions/lambdaexpressions/codegen/javagen/LambdaExpressionsJavaGenVisitor.java
SE-FDr: done in new version

monticore-grammar/src/main/java/de/monticore/codegen/javagen/JavaGenSymTypeRelations.java
SE-FDr: done in new version

monticore-grammar/src/main/java/de/monticore/codegen/javagen/typeconverter/JavaToStringConversionHandler.java
SE-FDr: removed in new version as it is MontiArc specific

PR #290, opened 2026-05-19
Add priority modifiers to DSTLs

luepges:


monticore-generator/src/main/java/de/monticore/dstlgen/grammartransformation/DSL2TransformationLanguageVisitor.java
```diff
@@ -207,6 +209,12 @@ public void visit(ASTExternalProd srcNode) {
     tfLang.addAbstractProd(productionFactory.createAbstractProd(srcNode));
   }
 
+  @Nullable protected ProdSymbol overriddenProd = null;
```
SE-FDr, line 212: hmm, move it up? and document. It is not local to a method to just ignore this

monticore-generator/src/main/java/de/monticore/dstlgen/grammartransformation/ProductionType.java
```diff
@@ -4,19 +4,29 @@
...
  * operation resulting in production rules.
- *
+ * Each mapped to a range of priority-modifiers
  */
 public enum ProductionType {
 
-    PATTERN("Pat"), LIST("List"), REPLACEMENT("Rep"), NEGATION("Neg"), OPTIONAL("Opt");
+  PATTERN("Pat", 50000),
+  LIST("List", 40000),
+  REPLACEMENT("Rep", 60000), // TODO: see MCQualifiedType / MCQualifiedType & document
```
SE-FDr, line 13: better todo note? maybe link directly to the issue here as well.

SE-FDr (changes requested):

PR #304, opened 2026-06-08
remove legacy interpreter

SE-FDr:
removes the legacy interpreter as it has been superseded.
This does remove the legacy Values and replaces them with MCValues.


SE-FDr:
Afterwards: https://git.rwth-aachen.de/monticore/montiarc/core/-/merge_requests/1550

jra-se (approved):

PR #332, opened 2026-07-17
Cleanup se-commons versions

luepges:
* split between previous and currently used version

(While the generator has the dependencies of the old runtime, the current runtime should use the current log version)

### DIFF
--- a/prepare-next-release/0001-use-next-snapshot.patch
+++ b/prepare-next-release/0001-use-next-snapshot.patch
@@ -8,44 +8,40 @@ diff --git a/gradle.properties b/gradle.properties
 index cb9de09ab..3fdc12ea2 100644
 --- a/gradle.properties
 +++ b/gradle.properties
-@@ -17,12 +17,12 @@ org.gradle.caching=true
+@@ -17,10 +17,10 @@ org.gradle.caching=true
  # org.gradle.caching.debug=true
  
  # versions used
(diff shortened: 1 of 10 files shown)


prepare-next-release/0001-use-next-snapshot.patch
```diff
@@ -8,44 +8,40 @@ diff --git a/gradle.properties b/gradle.properties
...
+@@ -17,10 +17,10 @@ org.gradle.caching=true
  # org.gradle.caching.debug=true
  
  # versions used
 -version=7.9.0-SNAPSHOT
 -previous_mc_version=7.8.0
+-previous_se_commons_version=7.8.0
+-previous_cd4a_version=7.8.0
 +version=7.10.0-SNAPSHOT
 +previous_mc_version=7.9.0-SNAPSHOT
```
SE-FDr, line 20: this will be set to 7.9.0 later?

luepges: For a "real" release, the previous versions will be set to the last released versions.  
For this "fake" release, the all previous versions are set to the last version (built during the previous MontiVerse job)

luepges: (in addition: This PR adds the line to the patch, that removes lines 17&18 - so yes)

SE-FDr (approved):

PR #334, opened 2026-07-17
default-typeId-as-constructor

SE-FDr:
* allow usage of type identifiers as constructor per default
* check Type4Ast invariant: No node is expression and type identifier


SE-FDr:
https://github.com/MontiCore/ocl/pull/14

ahellwig (approved):

PR #339, opened 2026-07-20
JavaGen: Use traverser for VariableDeclarator

TomBursch:


luepges (approved):

monticore-grammar/src/main/java/de/monticore/statements/mcvardeclarationstatements/codegen/javagen/MCVarDeclarationStatementsJavaGenVisitor.java
```diff
@@ -78,15 +78,20 @@ public void traverse(ASTLocalVariableDeclarationStatement node) {
...
-      if (varDeclarator.isPresentVariableInit()) {
-        getPrinter().print(" = ");
-        varDeclarator.getVariableInit().accept(getTraverser());
-      }
+      varDeclarator.accept(getTraverser());
       state.endStatement();
     }
   }
 
+  @Override
```
SE-FDr, line 86:
```suggestion
  // Note: The current usage of VariableDeclarator in MontiCore
  // should make this directly compatible with VariableDeclarators in Java.
  // Thus we can use create a direct mapping.
  @Override
```

PR #357, opened 2026-08-05
type compatibility for tuple, functions

SE-FDr:


SE-FDr:
No tests, as the behavior does (currently) not change,
but the messages are more fitting.

jra-se (approved):

PR #358, opened 2026-08-06
Update Documentation Landing Page

TomBursch:


docs/scripts/preprocessing.bat
```diff
@@ -0,0 +1,109 @@
+@echo off
```
luepges: I would rather have a sh script for both linux & windows to avoid having to maintain both.

luepges:
grafik

RIP my beloved moving wings :) 
Thanks!

luepges (approved):

PR #361, opened 2026-08-13
Mill State

luepges:


monticore-generator/src/main/java/de/monticore/codegen/cd2java/mill/CDMillDecorator.java
```diff
@@ -29,7 +31,7 @@ public CDMillDecorator(final GlobalExtensionManagement glex,
     this.millDecorator = millDecorator;
   }
 
-  public void decorate(final ASTCDCompilationUnit inputCD, ASTCDCompilationUnit decoratedCD) {
+  public void decorate(SymbolTableService symbolTableService, final ASTCDCompilationUnit inputCD, ASTCDCompilationUnit decoratedCD) {
```
SE-FDr, line 34: maybe have this argument at the end?
input/decorated exist always for a decorator, the symboltableservice "feels" more optional/less essential.

monticore-generator/src/main/java/de/monticore/codegen/cd2java/mill/MillDecorator.java
```diff
@@ -638,4 +645,31 @@ protected List<ASTCDMethod> getAttributeMethods(String attributeName, String att
...
+    this.replaceTemplate(EMPTY_BODY, snapshotMethod, new TemplateHookPoint("mill.SnapshotMethod", stateClassName, superSymbolList));
+    return snapshotMethod;
+  }
+
+  protected ASTCDMethod addLoadMethod(List<DiagramSymbol> superSymbolList) {
+    String stateClassName = symbolTableService.getCDName() + "MillState";
+    ASTMCType stateType = this.getMCTypeFacade().createQualifiedType(stateClassName);
+    ASTCDParameter stateParam = this.getCDParameterFacade().createParameter(stateType, "state");
+
+    ASTCDMethod loadMethod = this.getCDMethodFacade().createMethod(PUBLIC_STATIC.build(), "load", stateParam);
```
SE-FDr, line 666: I really don't think this should just be called "load".
Maybe have both methods have connected names, like `getState()/setState()` or (readState/loadState) or restoreState?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/mill/MillStateDecorator.java
```diff
@@ -0,0 +1,67 @@
...
+    String millStateClassName = symbolTableService.getCDName() + "MillState";
+
+    ASTCDClass stateClass = CD4AnalysisMill.cDClassBuilder()
+            .setModifier(PUBLIC.build())
+            .setName(millStateClassName)
+            .build();
+    this.replaceTemplate(JAVADOC, stateClass,
+            JavaDoc.of("This class holds a reference to the mill without the use of a static state.",
+                            "It is NOT a snapshot, i.e., later calls to the static mill affect this class's state.",
+                            "Note: TypeCheck3 statics are not yet stored.")
```
SE-FDr, line 46: ok, tbf, so are all other static states, not just TC3...

monticore-generator/src/main/java/de/monticore/codegen/cd2java/mill/MillStateDecorator.java
```diff
@@ -0,0 +1,67 @@
...
+  public ASTCDClass decorate(List<ASTCDPackage> packageList) {
+    String millStateClassName = symbolTableService.getCDName() + "MillState";
+
+    ASTCDClass stateClass = CD4AnalysisMill.cDClassBuilder()
+            .setModifier(PUBLIC.build())
+            .setName(millStateClassName)
+            .build();
+    this.replaceTemplate(JAVADOC, stateClass,
+            JavaDoc.of("This class holds a reference to the mill without the use of a static state.",
+                            "It is NOT a snapshot, i.e., later calls to the static mill affect this class's state.",
```
SE-FDr, line 45: if this is not a snapshot, why does it say to add the snapshot methods above?

monticore-generator/src/main/resources/mill/LoadMethod.ftl
```diff
@@ -0,0 +1,6 @@
+<#-- (c) https://github.com/MontiCore/monticore -->
+${tc.signature("superSymbolList")}
+mill = state.mill;
+<#list superSymbolList as superSymbol>
+${superSymbol.getFullName()?lower_case}.${superSymbol.getName()}Mill.initMe(state.${superSymbol.getName()?uncap_first}Mill);
```
SE-FDr, line 5: What partially annoys me, is, that this cannot? be used to set the Mills to the original state.
But I guess, that would work?:
```
AMill.init();
// do stuff
s = A.getState()
AMill.reset()
// init and do stuff with BMill
BMill.reset()
A.setState(s)
// happily use AMill again
```

SE-FDr: I think we are missing a lot of tests currently,
As there are quite some combinations of how languages are related to each other
and what to do to switch between them, like reset or no reset.

SE-FDr (changes requested):

monticore-generator/src/main/resources/mill/LoadMethod.ftl
SE-FDr: ok, just an experimental feature, tests can be added later.

PR #363, opened 2026-08-18
Deprecate the TypeDispatcher

jra-se:


SchmidtMarc:
Why do we deprecate it?

jra-se:
> Why do we deprecate it?

The original goal of modularity and extensibility requires double dispatch. However, the TypeDispatcher is only a single-dispatch mechanism and is therefore just a slower, more complicated, and more verbose form of an `InstanceOf`. Especially in light of newer Java features such as `InstanceOf` pattern matching and `switch` pattern matching, there are good (better!) alternatives.

Instead of the TypeDispatcher, you should therefore use either pattern matching for `InstanceOf` or `switch` if you are dealing with a complete language definition. In components that may be extended in the future, however, you should use the Visitor pattern.

SE-FDr:
> Instead of the TypeDispatcher, you should therefore use either pattern matching for InstanceOf or switch if you are dealing with a complete language definition. In components that may be extended in the future, however, you should use the Visitor pattern.

Read: use the Visitor pattern basically always with only VERY few exceptions.

monticore-generator/src/main/java/de/monticore/codegen/cd2java/typedispatcher/TypeDispatcherDecorator.java
```diff
@@ -66,6 +67,9 @@ public ASTCDClass decorate(ASTCDCompilationUnit input) {
 
     addGettersAndSetters(optAttributes, dispatcher);
 
+    replaceTemplate(ANNOTATIONS, dispatcher,
+        new StringHookPoint("@Deprecated(forRemoval = true)"));
```
SE-FDr, line 72: + comment to what to use instead?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/typedispatcher/TypeDispatcherDecorator.java
```diff
@@ -221,6 +225,10 @@ protected List<ASTCDMethod> createIsAsMethods(List<AttributeName> names) {
       replaceTemplate(EMPTY_BODY, isMethod, new TemplateHookPoint(
           "dispatcher.IsMethod", name.simpleName, name.parameterName));
 
+      replaceTemplate(ANNOTATIONS, isMethod,
+          new StringHookPoint("@Deprecated(forRemoval = true)"));
+
+
```
SE-FDr:
```suggestion
```

monticore-generator/src/main/java/de/monticore/codegen/cd2java/typedispatcher/TypeDispatcherDecorator.java
```diff
@@ -232,6 +240,10 @@ protected List<ASTCDMethod> createIsAsMethods(List<AttributeName> names) {
       replaceTemplate(EMPTY_BODY, asMethod, new TemplateHookPoint(
           "dispatcher.AsMethod", name.simpleName, name.parameterName));
 
+      replaceTemplate(ANNOTATIONS, asMethod,
+          new StringHookPoint("@Deprecated(forRemoval = true)"));
+
+
```
SE-FDr:
```suggestion
```

monticore-grammar/src/main/java/de/monticore/expressions/assignmentexpressions/cocos/LiteralAssignmentMatchesRegExExpressionCoCo.java
```diff
@@ -48,18 +45,10 @@ public void check(ASTAssignmentExpression node) {
...
-
       if (leftType.isRegExType() &&
-          expressionsDispatcher.isExpressionsBasisASTLiteralExpression(node.getRight())) {
-        ASTLiteralExpression literalExpression =
-            expressionsDispatcher.asExpressionsBasisASTLiteralExpression(node.getRight());
-        if (literalDispatcher.isMCCommonLiteralsASTStringLiteral(literalExpression.getLiteral())) {
-          String s = literalDispatcher.asMCCommonLiteralsASTStringLiteral(literalExpression.getLiteral())
-              .getSource();
+          node.getRight() instanceof ASTLiteralExpression literalExpression) {
+        if (literalExpression.getLiteral() instanceof ASTStringLiteral stringLiteral) {
```
SE-FDr: hmm.. connect with `&&` instead of having multiple if's? not really important, though, both is fine

monticore-grammar/src/main/java/de/monticore/interpreter/util/TypeDispatcherHotfix.java
```diff
@@ -16,25 +14,19 @@
 public class TypeDispatcherHotfix {
```
SE-FDr, line 11: should be deprecated and removed -> it is not needed anymore.

monticore-grammar/src/main/java/de/monticore/symbols/oosymbols/types3/OOSymbolsSymTypeRelations.java
```diff
@@ -103,11 +102,8 @@ protected Optional<SymTypeOfFunction> _getAbstractFunctionOfFunctionalInterFace(
               .flatMap(List::stream)
               .toList();
       List<SymTypeOfFunction> abstractMethods = methods.stream()
-          .filter(m -> OOSymbolsMill.typeDispatcher()
-              .isOOSymbolsMethod(m.getSymbol())
-          )
-          .filter(m -> OOSymbolsMill.typeDispatcher()
-              .asOOSymbolsMethod(m.getSymbol()).isIsAbstract()
+          .filter(
+              m -> m.getSymbol() instanceof MethodSymbol methodSymbol && methodSymbol.isIsAbstract()
```
SE-FDr: length

SE-FDr (changes requested):
Seems rather few changes, I expected more?

SE-FDr (approved):

jra-se:
Pipeline currently failes, as the JavaDoc references the SingleStepTraverser that is added in https://github.com/MontiCore/monticore/pull/366

PR #371, opened 2026-09-07
update class2mc to 7.9.1

SE-FDr:


luepges (approved):
(Required due to circular dependency of class2mc :( )

PR #372, opened 2026-09-07
SymTypeExpression and wrappers return a supplier in case the symbol c…

SE-FDr:
…an not be resolved immediatly


monticore-generator/src/main/java/de/monticore/codegen/cd2java/methods/mutator/MandatoryMutatorDecorator.java
```diff
@@ -25,13 +25,22 @@ public MandatoryMutatorDecorator(final GlobalExtensionManagement glex) {
...
   }
 
   protected ASTCDMethod createSetter(final ASTCDAttribute ast) {
-    String name = String.format(SET, StringUtils.capitalize(getDecorationHelper().getNativeAttributeName(ast.getName())));
-    ASTCDMethod method = this.getCDMethodFacade().createMethod(PUBLIC.build(), name, this.getCDParameterFacade().createParameters(ast));
-    this.replaceTemplate(EMPTY_BODY, method, new TemplateHookPoint("methods.Set", ast));
+    ASTCDAttribute attribute = ast;
+    String templateName = "methods.Set";
+    if (getDecorationHelper().isSupplier(ast.getMCType())) {
+      // expose the unwrapped type (Supplier<X> -> X) in the setter parameter; the Supplier stays hidden
```
SE-FDr, line 35: This may need another method:
one that allows one to set a supplier (for Lazy creation in, e.g., SymbolAdapters).

monticore-generator/src/main/java/de/monticore/codegen/cd2java/methods/AccessAsSupplierTyps.java
```diff
@@ -0,0 +1,22 @@
...
+
+  // TODO also accept non-fully-qualified spellings
+  public static final String[] SUPPLYLIST = {
+      "de.monticore.types.check.SymTypeExpression",
+      "java.util.List<de.monticore.types.check.SymTypeExpression>",
+      "java.util.Optional<de.monticore.types.check.SymTypeExpression>"
+  };
+
+
+  public static boolean shouldBeSupplied(String t) {
```
SE-FDr: `shouldBeSupplied` -> `shouldHaveSupplier`/`shouldBeAccessedViaSupplier`/etc.

monticore-generator/src/main/java/de/monticore/codegen/cd2java/methods/AccessAsSupplierTyps.java
```diff
@@ -0,0 +1,22 @@
+/* (c) https://github.com/MontiCore/monticore */
+package de.monticore.codegen.cd2java.methods;
+
+import java.util.Arrays;
+
+
+public class AccessAsSupplierTyps {
```
SE-FDr:
```suggestion
public class AccessAsSupplierTypes {
```

SE-FDr: needs documentation on what this class is for

monticore-generator/src/main/java/de/monticore/codegen/cd2java/methods/AccessAsSupplierTyps.java
```diff
@@ -0,0 +1,22 @@
...
+
+  public static final String SUPPLIER_TYPE = "java.util.function.Supplier";
```
SE-FDr: does this need to be public?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/methods/AccessAsSupplierTyps.java
```diff
@@ -0,0 +1,22 @@
...
+
+  // TODO also accept non-fully-qualified spellings
+  public static final String[] SUPPLYLIST = {
```
SE-FDr: does this need to be public?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
```diff
@@ -113,6 +117,25 @@ public boolean isOptional(ASTMCType type) {
     return false;
   }
 
+  public boolean isList(ASTMCType type) {
+    if (type instanceof ASTMCListType) {
```
SE-FDr: does isListType not work?
this may break easily, as it relies on instanceof, which should be avoided in close to all cases (for MC languages)

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
```diff
@@ -113,6 +117,25 @@ public boolean isOptional(ASTMCType type) {
...
+      return true;
+    } else if (type instanceof ASTMCGenericType) {
+      String simpleType = ((ASTMCGenericType) type).printWithoutTypeArguments();
+      return "List".equals(Names.getSimpleName(simpleType));
+    }
+    return false;
+  }
+
+  public boolean isSupplier(ASTMCType type) {
+    if (type instanceof ASTMCGenericType) {
```
SE-FDr: same as above

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
SE-FDr: additionally, we would not want two Methods doing the same thing

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
```diff
@@ -159,12 +182,34 @@ public String getSimpleNativeType(String nativeAttributeType) {
...
+      inner = "new java.util.ArrayList<>()";
+    } else if (isOptional(type)) {
+      inner = "Optional.empty()";
+    } else if (isSupplier) {
+      inner = "null";
+    } else {
+      return;
     }
+
+    String defaultValue;
```
SE-FDr, line 200: the `defaultValue` is what comes after the `=`.
This is more like the `defaultInitialization` or simmilar?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
```diff
@@ -181,7 +226,18 @@ public String getNativeAttributeName(String attributeName) {
 
   public ASTMCTypeArgument getReferenceTypeFromOptional(ASTMCType type) {
     Preconditions.checkArgument(isOptional(type));
-    return ((ASTMCGenericType) type).getMCTypeArgumentList().getFirst();
+    return ((ASTMCGenericType) type).getMCTypeArgumentList().get(0);
+  }
+
+  public ASTMCTypeArgument getReferenceTypeFromSupply(ASTMCType type) {
```
SE-FDr:
```suggestion
  public ASTMCTypeArgument getReferenceTypeOfSupplier(ASTMCType type) {
```
?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
```diff
@@ -181,7 +226,18 @@ public String getNativeAttributeName(String attributeName) {
...
+    Preconditions.checkArgument(isSupplier(type));
+    return ((ASTMCGenericType) type).getMCTypeArgumentList().get(0);
+  }
+
+  public ASTMCType createSupplierTypeOf(ASTMCType inner) {
```
SE-FDr, line 244: Would this fit more in the MCTypeFacade?

monticore-grammar/src/main/java/de/monticore/symbols/basicsymbols/_symboltable/TypeSymbol.java
```diff
@@ -21,11 +21,11 @@ public void setFunctionList(List<FunctionSymbol> methodList){
   }
 
   public List<SymTypeExpression> getSuperClassesOnly(){
-    return superTypes;
+    return superTypes.get();
```
SE-FDr: prefer something like `getSuperTypesList()`

monticore-grammar/src/main/java/de/monticore/symbols/basicsymbols/_symboltable/TypeSymbol.java
```diff
@@ -21,11 +21,11 @@ public void setFunctionList(List<FunctionSymbol> methodList){
...
   }
 
   public List<SymTypeExpression> getInterfaceList(){
-    return superTypes;
+    return superTypes.get();
```
SE-FDr: here as well

monticore-grammar/src/main/java/de/monticore/symbols/basicsymbols/_symboltable/TypeSymbol.java
Chrisslue: Why do both methods return the same? Is this correct?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
Chrisslue: There is also an isOptional(ASTMCType) function in DecorationHelper. Should I also remove that one? Since there were two here, I assumed it would be fine to do it like this.

monticore-generator/src/main/java/de/monticore/codegen/cd2java/methods/AccessAsSupplierTyps.java
Chrisslue: resolved

monticore-generator/src/main/java/de/monticore/codegen/cd2java/methods/AccessAsSupplierTyps.java
Chrisslue: resolved

Chrisslue: resolved

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
Chrisslue: I moved the "= " down so it is the defaulValue now

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
Chrisslue: resolved. I also did this rename for Optional

monticore-grammar/src/main/java/de/monticore/symbols/basicsymbols/_symboltable/TypeSymbol.java
Chrisslue: resolved

monticore-grammar/src/main/java/de/monticore/symbols/basicsymbols/_symboltable/TypeSymbol.java
Chrisslue: resolved

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
Chrisslue: I created a copy of the code into the facade, however, since generator seemingly uses a previous version of the grammar (implementation "de.monticore:monticore-grammar:$previous_mc_version"), it cant be used in this release. therefore I created both and we can migrate after. I still need to enter the correct version.

monticore-generator/src/main/java/de/monticore/codegen/cd2java/methods/AccessAsSupplierTyps.java
Chrisslue: see below. cant move it yet

monticore-runtime/src/main/java/de/monticore/symboltable/ClearingMemorizer.java
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Lazily applies the mapper to the value of the source, evaluated at most once.
+   * A null value of the source stays null. Nothing is evaluated until the result is requested,
+   * e.g. to copy a not-yet-resolvable attribute of a symbol without forcing it.
+   */
+  public static <T, R> Supplier<R> map(Supplier<T> source, Function<? super T, ? extends R> mapper) {
+    return new ClearingMemorizer<>(() -> {
+      T value = source.get();
+      return value == null ? null : mapper.apply(value);
```
SE-FDr: why can the mapper not accept null?

monticore-runtime/src/main/java/de/monticore/symboltable/ClearingMemorizer.java
```diff
@@ -0,0 +1,42 @@
...
+package de.monticore.symboltable;
+
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * Evaluates the wrapped supplier at most once and caches the result. Afterwards the
+ * supplier is released (and with it everything it captured, e.g. the JSON it deserializes from).
+ */
+public class ClearingMemorizer<T> implements Supplier<T> {
```
SE-FDr, line 11: Why is it called Clearing? that is an internal detail.

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
```diff
@@ -301,6 +301,24 @@ public ASTMCType unwrapSupplier(ASTMCType type) {
...
+   * so the internal wrapper never leaks into the public API.
+   */
+  public ASTMCType toPublicSupplierType(ASTMCType type) {
+    return createStdSupplierTypeOf(unwrapSupplier(type));
+  }
+
+  /**
+   * Wraps the type of every attribute that {@link #shouldHaveSupplier} selects into the internal supplier type.
+   */
+  public void wrapSuppliers(Collection<ASTCDAttribute> attributes) {
```
SE-FDr, line 316: wrapSuppliers does not say anything about storing the values

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
```diff
@@ -301,6 +301,24 @@ public ASTMCType unwrapSupplier(ASTMCType type) {
     return type;
   }
 
+  /**
+   * The type that generated code exposes for a supplied attribute:
+   * {@code __internal__Supplier<X>} (or an unwrapped X) becomes {@code java.util.function.Supplier<X>},
+   * so the internal wrapper never leaks into the public API.
+   */
+  public ASTMCType toPublicSupplierType(ASTMCType type) {
+    return createStdSupplierTypeOf(unwrapSupplier(type));
```
SE-FDr, line 310: This makes me wonder if the setup is too complex.
I am not sure.

monticore-grammar/src/main/java/de/monticore/types/check/SymTypeExpressionFactory.java
```diff
@@ -161,6 +161,36 @@ public static SymTypeOfObject createTypeObjectViaSurrogate(String name, IBasicSy
     return new SymTypeOfObject(typeSymbol);
   }
 
+
+  public static TypeSymbol resolveTypeSymbolOrLogError(
```
SE-FDr, line 165: This should not be part of its public api, it has nothing to do with the Factory directly

monticore-runtime/src/main/java/de/monticore/symboltable/ClearingMemorizer.java
Chrisslue: Is that an internal detail? Isn't this just saying what it does? Internal would be how it does it.
The "clearing" is supposed to communicate why someone would use it. To clear/release all data it holds after resolving.
What would you call it?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
Chrisslue: What do you mean storing?
Are you talking about the method mutating the attributes parameter?
Since the method returns void this would communicate it mutating the parameter in my mind.

If the function were to return the new attributes, we would copy everything, which I don't see a reason to

monticore-generator/src/main/java/de/monticore/codegen/cd2java/DecorationHelper.java
Chrisslue: What part is too complex? Do I have too many small functions? Or are you unhappy with the __internal__Supplier+ normal Supplier setup and just want to remove that part of the design?

monticore-generator/src/main/java/de/monticore/codegen/cd2java/methods/mutator/MandatoryMutatorDecorator.java
Chrisslue: I think this comment is out of date

monticore-runtime/src/main/java/de/monticore/symboltable/ClearingMemorizer.java
Chrisslue: True, i moved that decision to the outside.

monticore-grammar/src/main/java/de/monticore/symbols/basicsymbols/_symboltable/TypeSymbol.java
SE-FDr: ... yes, kind-of.
It is questionable design wise,
but the idea is that we cannot distinguish between them for TypeSymbols,
so we simply return them all.

PR #378, opened 2026-09-09
Fix Resolving StringIndexOutOfBoundsExceptions

luepges:
* Fix StringIndexOutOfBounds
* Fix eager getTypeInfo (in deprecated TC1)
* Improve error format

### DIFF
--- a/monticore-generator/src/main/resources/_symboltable/iartifactscope/GetRemainingNameForResolveDown.ftl
+++ b/monticore-generator/src/main/resources/_symboltable/iartifactscope/GetRemainingNameForResolveDown.ftl
@@ -2,7 +2,7 @@
   List<String> remainingSymbolNames = new ArrayList<>();
   String packageAS = this.getPackageName();
   if(symbolName.startsWith(packageAS)){
-    if(!packageAS.isEmpty()){
+    if(!packageAS.isEmpty() && symbolName.length() > packageAS.length()){
       symbolName = symbolName.substring(packageAS.length()+1);
     }
     String asName = this.getName() + ".";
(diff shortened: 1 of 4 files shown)


monticore-generator/src/main/resources/_symboltable/iartifactscope/GetRemainingNameForResolveDown.ftl
```diff
@@ -2,7 +2,7 @@
   List<String> remainingSymbolNames = new ArrayList<>();
   String packageAS = this.getPackageName();
   if(symbolName.startsWith(packageAS)){
-    if(!packageAS.isEmpty()){
+    if(!packageAS.isEmpty() && symbolName.length() > packageAS.length()){
```
SE-FDr: hmm, I am not sure if this is the cleanest way...

monticore-runtime/src/main/java/de/monticore/symboltable/IScope.java
```diff
@@ -83,7 +83,7 @@ public interface IScope {
   boolean isPresentName();
 
   default List<String> getRemainingNameForResolveDown(String symbolName) {
-    return Lists.newArrayList(isPresentName() && !getName().isEmpty() && symbolName.startsWith(getName()) ? symbolName.substring(getName().length() + 1) : symbolName);
+    return Lists.newArrayList(isPresentName() && !getName().isEmpty() && symbolName.startsWith(getName()) && symbolName.length() > getName().length() ? symbolName.substring(getName().length() + 1) : symbolName);
```
SE-FDr:
```suggestion
    return Lists.newArrayList(
        isPresentName()
        && !getName().isEmpty()
        && symbolName.startsWith(getName())
        && symbolName.length() > getName().length()
        ? symbolName.substring(getName().length() + 1)
        : symbolName
    );
```
? readability is getting really hard these days

monticore-generator/src/main/resources/_symboltable/iartifactscope/GetRemainingNameForResolveDown.ftl
SE-FDr: since it already has a startsWith check...
```suggestion
    if(!packageAS.isEmpty() && !symbolName.equals(packageAS)){
```

SE-FDr (approved):

